PR #3, opened 2020-10-01
marketo destination blog post

bleonard:
screencapture-localhost-3001-blog-marketo-destination-2020-10-01-13_27_06


andyjih (commented):
Do the screenshots need to be obfuscated?

bleonard:
@andyjih good catch! Well, it would have been but those are all fake numbers.

PR #9, opened 2020-10-09
Integrations page

andyjih:


components/home/integrationCard.tsx
```diff
@@ -0,0 +1,88 @@
+import { Card, Badge } from "react-bootstrap";
+import SmallIntegration from "./smallIntegration";
+import Link from "next/link";
+import { formatNameIntoLink } from "../../utils/formatters";
+
+export default function IntegrationCard({
```
evantahler, line 5: There's something wonky about the corners of the card
Screen Shot 2020-10-09 at 3 14 46 PM

components/home/integrationCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+  showLink = true,
+}: {
+  name: string;
+  image: string;
+  slug: string;
+  type?: string;
+  otherType?: string;
+  category?: string;
+  tag?: string;
+  showLink?: boolean;
```
evantahler, line 22: Should we display something visual on the card if there is another page to learn more?

components/home/integrationDestinations.tsx
```diff
@@ -5,24 +5,28 @@ export default function IntegrationDestinations() {
     <>
```
evantahler, line 7: This can be determined programmatically now!   You can take the top N destinations by priority and display them

components/home/integrationsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <Container>
+        <Row>
+          <Col>
+            <h2>Integrations with all of your tools</h2>
+            <br />
+          </Col>
+        </Row>
+
+        <Row>
+          <Col lg={4} md={4} sm={6} xs={6}>
```
evantahler: Another place to use the top N by score?

components/home/integrationSources.tsx
```diff
@@ -3,13 +3,33 @@ import IntegrationButton from "../home/integrationButton";
 export default function IntegrationSources() {
```
evantahler, line 5: Another place to use the top N by score?

data/plugins.ts
```diff
@@ -0,0 +1,161 @@
+export const PluginData = [
```
evantahler, line 1: ❣️

evantahler: Future story: get this list from NPM or the monorepo

pages/integrations.tsx
```diff
@@ -0,0 +1,108 @@
...
+        style={{ textAlign: "center", paddingTop: 100, paddingBottom: 100 }}
+      >
+        <Container>
+          <Row>
+            <Col>
+              <h1>
+                Connect your data stack with all the tools that your team uses
+              </h1>
+              <p className="text-big-muted">
+                Grouparoo is adding new Integrations all the time!
```
evantahler: This sub-heading is missing something for me.  Some Ideas:
* Grouparoo lets you import and export your data to the world's most popular tools
* Grouparoo integrates with the tools you need to get work done

evantahler (commented):
Looking good!

A few general comments:

* It looks like the `/destinations/*` and `/sources/*` pages are all pretty similar.  Do you want to templatize them and just load the data from our json file?  This way, there's only one layout to worry about as you change things/add a new one
* there's something that bothers me about the colors on the pills on `/integrations`.  The colors seem to have meaning (source's color !== destination's), but they aren't the same per-destination.  I would imagine the colors for each pill are always the same (source=red, destination=blue, etc)

Screen Shot 2020-10-09 at 3 25 05 PM

andyjih:
Closes T-598

andyjih:
@evantahler Thanks for the review! Made some updates based on your recommendation. Thoughts?

bleonard:
There's some way to make the cards in the same row the same height. 
I think that would make things look better.

bleonard:
Screen Shot 2020-10-09 at 5 04 26 PM
Learn more looks a bit funny to me. Maybe a new line or not there at all?

andyjih:
@bleonard Yeah, removed the "Learn More". I tried to make the cards the same height but got stuck. If you have any pointers, that'd be great!

bleonard:
here's the example of same height. https://react-bootstrap.github.io/components/cards/
Maybe it's putting them in a `<CardDeck>` ?

andyjih:
Added a hover state to clickable cards. I'll punt on the height thing for now.

PR #24, opened 2020-10-16
Pricing page

andyjih:


evantahler (approved):
Looking good!

I'm generally skeptical of Google Analytics 1) being effective at all any more and 2) storing GET url params... but that's not a reason to hold up this PR!

evantahler:
(I should add specifically because with Next, we don't actually 'change pages' in the normal way sometimes)

PR #25, opened 2020-10-16
Using Typescript to create a Robust API between your frontend and backend

evantahler:
screencapture-localhost-3001-blog-typescript-frontend-backend-2020-10-16-15_46_24


pages/blog/typescript-frontend-backend.mdx
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  if (time === 0) return <div>loading...</div>;
+
+  const formattedTime = new Date(time).toLocaleString();
+  return <div>The time is: {formattedTime}</div>;
+}
+```
+
+Now we have enforced that the type of `response` in the `load()` method above will match the Action, being `{ time: number; }`. We will now get help from Typescript if don’t use that response value as a number, if we try to assign it to a string variable or otherwise use it improperly.
```
andyjih: "We will now get help from Typescript if don’t use that response value as a number, if we try to assign it to a string variable or otherwise use it improperly." 

Too many "if"s here? Can you take another pass at this?

andyjih (changes requested):
Just one minor thing, but looks good! 👍

evantahler:
Thanks @andyjih!

PR #26, opened 2020-10-17
Migrate to `next-mdx-remote` and simplify MDX Pages

evantahler:
`next-mdx-enhanced` was pretty janky.  We had a lot of custom code that was not "the Next way of doing things", and rather slow to build.  This PR:
- [x] replace `next-mdx-enhanced` with [`next-mdx-remote`](https://github.com/hashicorp/next-mdx-remote) 
- [x] remove `blogEngine`, replace with `getStaticRoutes` + `getStaticProps`
- [x] remove `docsEngine`, replace with `getStaticRoutes` + `getStaticProps`
- [x] remove `getSitemap` and simplify sitemap generation with a glob
- [x] Re-type the collection of blogPosts and docPages
- [x] refactor the loader code used in `getStaticProps` to be shared and simpler to read
- [x] customize the md/x renderer to open all external links in a new tab

Side Effects:
* `npm run build` is ~4x faster
* All new pages are added to the sitemap automatically (we glob `/pages` and find everything)
* There's no special files needed to "trick" mdx/typescript into loading the collection of blogposts/docs that we need to change on each boot
* We now properly use Next's static site generation so JS/bundles can be shared and chunked across pages (faster load times for viewers)

Screen Shot 2020-10-16 at 10 42 51 PM

Notes:
* `next-mdx-remote`  now requires that MDX files no longer `import`/`require` anything directly.  Components can still be used with JSX, needed to be loaded into the rendered ahead of time, and will be available for all MDX pages of the same type (blog, docs)

Closes T-634


evantahler:
https://github.com/grouparoo/www.grouparoo.com/pull/26/commits/552ce538d83f721494d9ecb8728008d3c7a45770 closes T-633

PR #35, opened 2020-10-25
Adds /whats-new page

bleonard:
This also adds an RSS Feed and a [JSON Feed](https://jsonfeed.org/) for the new section as well as the blog.

Screen Shot 2020-10-25 at 2 16 36 AM

Screen Shot 2020-10-25 at 2 17 14 AM


evantahler:
I really like where this is going!

* Perhaps the links to Github should always be version numbers (instead of PRs).  I think this will help with sharing a clearer CTA to "upgrade/download now".  Of course, you can link to PRs in the body if they are interesting.  
* This also sort of implies that each post should be more directly linked to a version number maybe?

bleonard:
I like that. I'm trying to collect meta data on these posts. Linking to github somehow shows a full cycle which I think is valuable.

"version" seems like a good one too. what's the best way to backport these?

evantahler:
Starting from the assumption that ever "real" release (skipping the alphas) has something worth talking about, I would suggest refactoring this so that there's a .mdx for each release (from https://github.com/grouparoo/grouparoo/releases).  Perhaps the release is the file name?

evantahler:
Oh, please add a link to this page in the docs/community menu!

bleonard:
In general, I don't think these entries are 1-to-1 with the npm releases. Some of those are bigger and some are smaller. I was going for more of a feed of things to try out and product updates. That might be zero for a npm release and it might be more than 1.

PR #36, opened 2020-10-25
Disable _next listing when deployed

evantahler:
Looks like `/_next` lists all of our assets.  Not that there's anything secret in there, but it's annoying.  We can use `vercel.json` to redirect visitors to these index routes somewhere else.  We can't know the full routes of all the data in `_next/data`, but since these are etag-like slugs, they will be reasonably hard to guess. 

Thanks to https://stackoverflow.com/questions/62614474/how-to-disable-directory-listing-in-next-js-deployed-to-vercel

Screen Shot 2020-10-25 at 2 03 50 PM

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,0 +1,44 @@
+{
+  "redirects": [
+    {
+      "source": "/_next",
+      "destination": "/404",
+      "statusCode": 404
+    },
+    {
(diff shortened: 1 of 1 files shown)


evantahler:
Thanks for finding this @bleonard!

PR #44, opened 2020-10-27
whats-new get their own urls

bleonard:
This is better for social sharing because we can give metadata and other such things.

for example: https://www.grouparoo.com/whats-new/2020-04-23-sailthru


evantahler:
Nice! 
Is there a way to get to those per-item URLs from the list view?  Perhaps a `permalink` or somesuch?

Also, perhaps the generated files (sitemap, etc) shouldn't be checked in / Git-Ignored?

bleonard:
You mean, make them clickable from the list view? Yes, there is a way! `<a>` tags would work.
I decided not to do that because it hints that there is more content. But it's the same. Think it should link?

I went back and forth related to build things and, in the end, decided it was better for visibility and deployment to check them in and have them built from [github action](https://github.com/grouparoo/www.grouparoo.com/pull/44/commits/1b46e8c9d63c15852d77f56b9ae6c4d79e66d57d).

PR #45, opened 2020-10-27
How we work blogpost

andyjih:


evantahler:
FYI your tags are both black because they aren't defined yet in [`badgeTypes`](https://github.com/grouparoo/www.grouparoo.com/blob/master/utils/blogTags.tsx#L5-L11).  Without that, they also won't appear as a category, `/blog/tag/x` either.

Screen Shot 2020-10-27 at 7 04 11 PM

evantahler:
> and none of us had any intention or desire to move, we decided to make the company remote-first. 

Seems like a soft point.  We didn't choose it because we thought it was better for some reason?

> Default to open

Interesting! To me that phrase means that we "work in the open" (either in public codebases or at least docs the rest of the team has access too).  Perhaps for clarity you mean "Open source first" or something like that?

Also, thanks for gathering our team pictures for KeyValues :D

evantahler:
> One of the many cultural practices that I loved from my time at Stripe was how nearly every decision or idea was written in a document. 

... And we do this too! (You never actually said that this is what we do)

andyjih:
@evantahler thanks for the comments. I made some updates!

andyjih:
@evantahler updated, and ready for re-review

evantahler (approved):

PR #46, opened 2020-10-28
Snowflake and Salesforce integrations pages

andyjih:


andyjih:
Closes T-663

pages/integrations/destinations/salesforce.tsx
```diff
@@ -0,0 +1,226 @@
...
+      </Container>
+      <div
+        id="Salesforce Use Cases"
+        className="bg-dark text-white"
+        style={{ paddingTop: 50, paddingBottom: 50 }}
+      >
+        <Container>
+          <Row>
+            <Col lg={6} md={6} xs={12}>
+              <h2>Grouparoo & Salesforce</h2>
```
bleonard: I'm not sure but this might be right with `&amp;` instead of `&`

bleonard: Or put it in a JS string: `{"Grouparoo & Salesforce"}`

bleonard (approved):

PR #80, opened 2020-11-18
About page updates

andyjih:


evantahler (approved):
👍 

Looks good!  Not a blocker - but since some of those logos have white backgrounds and others don't, you might try putting them all in boxes with white backgrounds to be more uniform and enforce a spacial layout given their differing aspect ratios.  Sort of like we do on the list of integrations

PR #92, opened 2020-11-23
Update to Next 10.0.2 and Allow embded img alt tags

evantahler:


__tests__/integration/html.ts
```diff
@@ -79,9 +79,7 @@ describe("sitemap integration", () => {
           }
 
           const alt = tag.attr("alt");
-          if (!alt) {
-            altTagMissing.push(name);
-          }
+          if (!alt && !name.match(/^data:image/)) altTagMissing.push(name);
```
bleonard: Why shouldn't `data:image` have an `alt` name for SEO?

evantahler: It's a Next thing I don't quite understand - I think they are making a placeholder image (transparent gif) that displays until our real image loads?  I could cross-check this with `aria-hidden="true"` perhaps

Screen Shot 2020-11-23 at 10 38 19 AM

bleonard: yeah. or maybe that role=presentation
never seen that before so it's probably safe

evantahler: Yeah! That works (see https://github.com/grouparoo/www.grouparoo.com/pull/92/commits/99cbd22ee30ce8bcc68931ec8decdf2558e41f7d)

PR #117, opened 2020-12-18
Updated hero image and homepage

andyjih:


pages/index.tsx
```diff
@@ -337,10 +338,10 @@ export default function IndexPage() {
 
       <br />
       <CustomerTestimonials />
-      <br />
 
       <div
         id="how-grouparoo-works"
+        style={{ backgroundImage: "linear-gradient(#242436,#43435F)" }}
```
evantahler, line 386: getting fancy!

evantahler:
@andyjih there's something wacky about the way you made that SVG.  You've embedded PNG data into the svg... which makes the file really big, but also it renders using canvas and looks kind of crazy.  It's also not resizable. Basically, you need to make sure that the logos you are using are also SVGs. 

Screen Shot 2020-12-17 at 5 25 24 PM

bleonard:
Wouldn't hurt to update screenshots at some point

PR #143, opened 2021-01-07
Update docs/plugins

evantahler:


pages/docs/development/plugins.mdx
```diff
@@ -160,7 +174,7 @@ An initializer is an Actionhero file which lives in `/src/initializers`. There a
 Learn more about Actionhero initializers [here](https://www.actionherojs.com/tutorials/initializers).
 
 Within initializers, you can register routes and create apps and connections for Grouparoo.
-If your Plugin is providing apps or connections you can use the `plugin.registerPlugin` method. See the example below.
+If your Plugin is providing apps or connections you can use the `plugin.registerPlugin` method. See the simplified example below. You can see the full file from [`@grouparoo/postgres` here](https://github.com/grouparoo/grouparoo/blob/master/plugins/%40grouparoo/postgres/src/initializers/plugin.ts)
```
bleonard, line 177: random not on github links. I've been doing a `sha` version like https://github.com/grouparoo/grouparoo/blob/c525ba5a5bcc1dfbab848991ea0475fa7fd5cb86/plugins/@grouparoo/postgres/src/initializers/plugin.ts so that it doesn't change out from under me, especially when linking specific lines. Think that's correct? Or is up-to-date-ness more important?

evantahler: I think it depends on what you want to share.  This link was for the "current version of the thing", so I wanted to use master.  But, in PRs or Issues, I think it makes sense to use the SHA.  Ideally, links from www would point to the 'current version' as that's what customers are using maybe?

PR #153, opened 2021-01-14
Signup form for slack community.

bleonard:


evantahler:
This seems like it might be a bit confusing with `/docs/community`... perhaps this is `/slack`?  We also have a DNS redirect from `community.grouparoo.com` to `/docs/community` which seems related

evantahler:
Also, I'd suggest adding this to the `/docs/community` sidebar and somewhere in the body content of `/docs/community`

PR #166, opened 2021-01-21
Add Sean to /about

seancdavis:
This also adjusts the layout a bit. The grid is four-up on large screens, three-up on medium, and stays the same on smaller screens.

I also used my white background photo because the blue was distracting among the rest of the lot.

---

Closes T-950


README.md
```diff
@@ -20,6 +20,12 @@ npm install
 npm run dev
 ```
 
+If you run into an issue installing Sharp on an Apple M1 machine, make sure `vips` is installed properly. ([See this issue](https://github.com/lovell/sharp/issues/2460#issuecomment-751491241) for more information.)
+
+```bash
+brew install vips
+```
+
```
evantahler, line 28: Good find!

evantahler (approved):
Great work! 


* ✅ The Grid changes make sense to me!
* ✅ You closed the Height Task
* ✅ Since this is a layout-only change, I don't believe that a test is needed

PR #176, opened 2021-01-29
Add property rule operations dictionary to Create Groups doc

seancdavis:
This takes the JSON output from grouparoo/grouparoo#1241 and presents it as a tabbed series of tables to represent the operations available depending on the type of property and underlying Grouparoo database.

It assumes there is some mechanism in place for updating these two new files in the `data` directory. [This is a manual process for now](https://github.com/grouparoo/grouparoo/pull/1241#pullrequestreview-578840071).

I wanted to do more with this but decided to wrap it just to get it out the door so I can move on to reorganizing the docs as a whole. In particular:

- I kept the styling simple just to look presentable/legible.
- I wanted to add anchors to headings so we could link directly to them. I ran into some trouble. I will explore as a separate task during docs cleanup.

### DIFF
--- a/components/docs/index.tsx
+++ b/components/docs/index.tsx
@@ -1,0 +1,5 @@
+import Image from "./image";
+import HavingProblems from "./havingProblems";
+import RuleOpsTable from "./ruleOpsTable";
+
+export { Image, HavingProblems, RuleOpsTable };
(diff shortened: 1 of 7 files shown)


evantahler (approved):
👍 

The tabs and expanding sections are really nice.

There's a story for adding anchor links to headers here - https://www.pivotaltracker.com/story/show/176694618.. Feel free to tackle it now or as part of the new `www/docs` work

PR #180, opened 2021-02-01
Next v10.0.6 + rebuild package-lock.json

evantahler:


evantahler:
Note for the future - It seems that this newer version of next changed... something (SASS dependency?) which no longer plays well with Prism (the CSS library we use for code highlighting).  I've moved Prism's CSS out of the sass compiler to a direct link for now...

PR #181, opened 2021-02-01
[Blost Post] Node.js Memory Error on Apple M1

seancdavis:


evantahler:
Nice start!

1. All posts need an image for SEO and the RSS feed
2. Perhaps mention that you can run older versions of node in compatibility mode? You need to run the terminal in Rosetta and re-nvm with it... https://stackoverflow.com/questions/64899827/nodejs-on-apple-silicon-m1

evantahler:
And there's still something wacky with your author image
Screen Shot 2021-02-01 at 1 18 33 PM

seancdavis:
NOTE: I'm pausing on this. Still having computer issues, so I don't feel like I've actually solved the problem that the post is presenting.

bleonard:
Look at those ethics. It not quite working never stopped anyone from blogging a solution! :-P

seancdavis:
Finally feel like I have enough info that I could update this into something publishable.

evantahler (approved):
👍 

@bleonard created the category of "notes" to be exactly this - Short, simple "field notes" on how to do something you learned.

PR #184, opened 2021-02-02
Rearrange Docs Taxonomy

seancdavis:
Rearranges our doc pages according to [this document](https://docs.google.com/document/d/1iOzxhJzRdnL4lnkNwb1Daqld1HXjy9ORU8CZmkgwm4c/edit#). This puts us in a place where, once merged, we can start to build out the missing pages and the pages that should be updated.

Items to note within the code:

- The sidebar navigation is now declarative. It is controlled in `data/docs-nav.ts`.
- Nav theoretically now supports an infinite number of levels, but I'm not using any more than we were previously.
- I created a separate story to make the breadcrumbs a little smarter and to be able to remove the `section` property within each doc.

And items to note within the content:

- 302 redirects are in place for any content that moved. These are listed in `data/docs-redirects.js`. They are broken up by section so that it's easier to find items. I'm making the assumption that some of these may change again while the docs section is in flux. So they are all temporary for now.
- There is a redirect in the main next config file for /support. That page seemed like it fit in-line with the rest of our docs so I combined it with another support section from another page and have the resulting page living in the docs section. I also made this a 302 in case we want to go back to /support at some point in the future.
- Everything is at least _close_ to where it should be. There are a few items to note:
    - /docs/config/settings is an empty placeholder file
    - /docs/running-grouparoo is an empty placeholder file
    - /docs/config/code-config is in the same state it was, but it should go away when the rest of the content in the config section gets overhauled.
    - same goes for /docs/config/first-time-setup
- I changed a few things from our original proposal. It's most noticeable in the getting started section. We have a story to give that section some love, though, so I think we're okay for now.


seancdavis:
A few lingering tasks:

- [x] Find a place for the three existing pages
- [x] Add redirects
- [x] Check if `section` property on each doc is doing anything and remove if not
- [x] Put in placeholders for (some _necessary_) pages that don't yet exist

__tests__/utils/inflectors.ts
```diff
@@ -0,0 +1,16 @@
+import { titleize } from "../../utils/inflectors";
```
evantahler, line 1: That's what these are called!

pages/docs/config/apps.mdx
```diff
@@ -1,7 +1,7 @@
 ---
-title: "Add Apps"
+title: "Apps"
 date: "2020-08-01"
-section: "guides"
+section: "config"
```
evantahler, line 4: Do the `sections` mater any more since the nav is built up manually?

pages/docs/config/permissions.mdx
```diff
@@ -0,0 +1,19 @@
...
+<Image
+  centered
+  source="Grouparoo Permissions"
+  alt="Grouparoo Permissions"
+  src="docs/deployment/api-key-permissions.png"
+  width={820}
+  height={646}
+/>
+
+The Community edition is configured for a single user, who has full control over the application.
```
evantahler: Actually, at the moment, we still have the notion of teams.  I'll add a story to discuss this!

pages/docs/deployment.mdx
```diff
@@ -31,115 +31,19 @@ curl -L https://www.grouparoo.com/docker-compose --output docker-compose.yml
...
-createdb grouparoo_development # Create a database for Grouparoo
-```
-
-### Step 2: Generate `package.json` and `.env`
+Continue with installation details in our Docker installation guide.
 
-In a new directory, run the following command to generate the files Grouparoo needs:
+<Button variant="outline-primary" href="/docs/deployment/docker">
+  Get Started with Docker
+</Button>
```
evantahler, line 38: I think this is OK for simplicity.  Next would want you to do something like `<Button onClick={router.push("/docs/deployment/docker")}/>` for the fancy hot-reloading page changes.

pages/docs/getting-started.mdx
```diff
@@ -0,0 +1,36 @@
...
+[![Deploy to Heroku](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/grouparoo/app-example)
+
+### Run locally with Docker & Docker Compose
+
+If you have [Docker](https://www.docker.com/) installed, you can also run Grouparoo locally:
+
+    # in a new directory
+    curl -L https://www.grouparoo.com/docker-compose --output docker-compose.yml
+    docker-compose up
+
```
evantahler, line 27: Now that you don't need redis or postgres to try out the app locally, do we want to keep the NPM way of trying out Grouparoo?

```
npx grouparoo init .
npm start
```

data/docs-nav.ts
```diff
@@ -0,0 +1,129 @@
...
+        title: "Properties",
+        path: "/docs/config/properties",
+      },
+      { title: "Groups", path: "/docs/config/groups" },
+      {
+        title: "Destinations",
+        path: "/docs/config/destinations",
+      },
+      {
+        title: "Settings (Coming Soon)",
```
evantahler: I think you can remove `Coming Soon` from the nav

data/docs-nav.ts
```diff
@@ -0,0 +1,129 @@
...
+      { title: "Environment", path: "/docs/config/environment" },
+      {
+        title: "Secrets",
+        path: "/docs/config/secrets",
+      },
+    ],
+  },
+
+  // -- Running Grouparoo --
+  { title: "Running Grouparoo (Coming Soon)", path: "/docs/running-grouparoo" },
```
evantahler: Same note about `Coming Soon`

data/docs-nav.ts
```diff
@@ -0,0 +1,129 @@
...
+  {
+    title: "Community",
+    path: "/docs/community",
+    children: [
+      { title: "Code of Conduct", path: "/docs/community/code-of-conduct" },
+      { title: "Contributions", path: "/docs/community/contributions" },
+      {
+        title: "Forum",
+        path: "https://github.com/grouparoo/grouparoo/discussions",
+      },
```
evantahler, line 97: Can `forum` and `roadmap` open in `_blank`?

evantahler (commented):
Overall, nice work!

**General**
- Manually declaring the nav seems OK to me! 
- There's something wacky with rendering code blocks.  On master, I had to move the CSS file for Prism out of the SCSS.  But... this is only on some pages (`/docs/getting-started` but not `docs/deployment`)

Screen Shot 2021-02-02 at 1 12 35 PM

- It used to be the case that the section headers were highlighted orange in the nav when on those pages (ie: `/docs/getting-started`) which seems to be lost now

pages/docs/config/apps.mdx
seancdavis: They are being used by the breadcrumbs. I'm currently working a story to rework the breadcrumbs and that should enable me to remove all of these.

pages/docs/config/permissions.mdx
seancdavis: Okay. I'm going to get rid of these mentions and we can address following the new story.

pages/docs/deployment.mdx
seancdavis: [I added a story to capture this](https://www.pivotaltracker.com/n/projects/2401683/stories/176786039).

data/docs-nav.ts
seancdavis: [Broke this one out separately, too](https://www.pivotaltracker.com/n/projects/2401683/stories/176786564).

PR #187, opened 2021-02-03
Add tabs to getting started page

seancdavis:
I played around with this for a bit. It's not a pretty solution. Curious if you have any thoughts on cleaning up the code, or if you're fine with this?


evantahler (approved):
Mixing HTML, React, and Markdown will always be ugly... but it's powerful!

PR #189, opened 2021-02-04
Support opening docs nav items in a new tab

seancdavis:
Let's try that again ... 😅


evantahler:
Looks great for `forum`.  `roadmap` should also open in a new tab.

PR #190, opened 2021-02-04
Update stale refs to docs pages

seancdavis:
There will also be a PR for the product, but this doesn't rely on it.


evantahler (approved):
Thank you!

PR #199, opened 2021-02-08
Add notes on option to use fake redis and sqlite

seancdavis:
I focused on the local Node.js page. I also saw references to them on the Environment page.

Should we mention them elsewhere, too?

### DIFF
--- a/pages/docs/installation/node.mdx
+++ b/pages/docs/installation/node.mdx
@@ -7,11 +7,18 @@ pullQuote: "This guide will help you run the Grouparoo application on your local
 
 This guide will help you run the Grouparoo application on your local machine using Node.js.
 
-### Step 1: Install Runtime and Databases
+## Step 1: Install Runtime and Databases
 
-Grouparoo is a [node.js](https://nodejs.org/) application and requires Node.js version 12 or higher. You can download Node.js from their [website](https://nodejs.org/) or through the package manager for your operating system. The same is true for Postgres and Redis.
+Grouparoo is a [node.js](https://nodejs.org/) application and requires Node.js version 12 or higher. You can download Node.js from their [website](https://nodejs.org/) or through the package manager for your operating system.
 
-We also need to create a database within Postgres for Grouparoo to use. Let's use one called `grouparoo_development`.
+By default, running Grouparoo locally relies on the following dependencies:
+
+- [Redis](https://redis.io/)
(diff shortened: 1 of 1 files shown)


pages/docs/installation/node.mdx
```diff
@@ -7,11 +7,18 @@ pullQuote: "This guide will help you run the Grouparoo application on your local
...
 This guide will help you run the Grouparoo application on your local machine using Node.js.
 
-### Step 1: Install Runtime and Databases
+## Step 1: Install Runtime and Databases
 
-Grouparoo is a [node.js](https://nodejs.org/) application and requires Node.js version 12 or higher. You can download Node.js from their [website](https://nodejs.org/) or through the package manager for your operating system. The same is true for Postgres and Redis.
+Grouparoo is a [node.js](https://nodejs.org/) application and requires Node.js version 12 or higher. You can download Node.js from their [website](https://nodejs.org/) or through the package manager for your operating system.
 
-We also need to create a database within Postgres for Grouparoo to use. Let's use one called `grouparoo_development`.
+By default, running Grouparoo locally relies on the following dependencies:
```
evantahler, line 14: I think the default (ie: what `grouparoo init` sets up for you) is to use mock redis and sqlite.

evantahler: We should advise them to change these defaults when they are ready to persist data

pages/docs/installation/node.mdx
```diff
@@ -40,10 +46,17 @@ nvm use v12 # use v12 of node.js
 sudo apt install redis-server # Install Redis
 
 sudo apt install postgresql postgresql-contrib # Install Postgres
+```
+
+### Create Grouparoo Database
+
+Using either Postgres or SQLite, create a database for Grouparoo to use. This is where the application data will be stored, such as profiles, apps, and sources. In our example, we've called the database `grouparoo_development`. This is what it looks like for Postgres:
```
evantahler, line 53: In SQLite, you don't "create" a database like you do with postgres.  There is no `createdb` equivalent in SQLlite

evantahler (changes requested):
The changes are good, but we should match the "defaults" to what `grouparoo init` does

evantahler:
I've rolled these changes into https://github.com/grouparoo/www.grouparoo.com/pull/203

PR #221, opened 2021-02-16
In-Page Navigation

evantahler:
This PR adds an in-page navigation for Docs and Blog Posts.  

Screen Shot 2021-02-15 at 4 38 57 PM
Screen Shot 2021-02-15 at 4 53 25 PM

The nav is sticky once you scroll down far enough and stays with you (if you have a medium-sized or larger viewport)

Screen Shot 2021-02-15 at 4 39 03 PM

* Re-layout the blog header
* Collapse the docs/navigation sections that are not active to save space and be less intimidating
* If there would be only 1 item in the page navigation, it's hidden.


components/docs/tableOfContents.tsx
```diff
@@ -51,14 +51,14 @@ export function TableOfContents() {
         <Card.Body>
           <small className="docsTableOfContents">
             {navItems.map((section, idx) => {
+              const isActiveSection = router.asPath.includes(section.path);
```
seancdavis, line 54: I like cleaning stuff like this up! We're using very similar logic above in the NavListItem component. Maybe we can share the logic?

seancdavis (approved):
Love this! I think it's totally fine as is. Here are some thoughts/questions:

- Should the component be in a shared or more generic location (outside of `components/docs`) since it's not only serving the docs?
- It would be cool if it could follow your progress like it does on Actionhero, but I don't think it's necessary. The important part of this work is here.

evantahler:
Thanks @seancdavis! I've moved the component somewhere more common via  https://github.com/grouparoo/www.grouparoo.com/pull/221/commits/e558e77d834f57b85d8d9e258c445df47cb24b5b.  

Following your progress would be cool, but I figured we would leave that for anther rev.  The way I did it on www.actionherojs.com is actually kind of buggy, now that I'm looking at it again.  If the sections between each header are short, it highlights 2 sections at once, so I'd want to put some more thought into it

Screen Shot 2021-02-16 at 8 26 01 AM

andyjih (approved):
woot

PR #226, opened 2021-02-17
Remove references to SCHEDULER

seancdavis:
This is to support changes in grouparoo/grouparoo#1314.

⚠️ Note that this could cause an issue for anyone who deploys between merging this and releasing the next core version (the one that will have this adjustment).


evantahler (approved):
Nice! 

Let's not merge this in until the next release.

PR #228, opened 2021-02-17
Wrap all MDX headers in links

evantahler:
Taking a page from the Stripe docs (https://stripe.com/docs/api/payment_intents/update), now all headers generated by MDX will be links so that you can click on them and the link + anchor in your address bar.  This applies to both `/docs` and `/blog`

Visual style is still the same until you hover:

Screen Shot 2021-02-17 at 10 30 28 AMScreen Shot 2021-02-17 at 10 30 29 AM


andyjih (approved):
Awesome! Thank you!

PR #232, opened 2021-02-18
Add support for overriding twitter card type

seancdavis:
This provides the option to override the the `twitter:card` meta tag, which is set to `summary` by default.

In frontmatter for a post, you could do:

```mdx
---
# ...
twitter_card: summary_large_image
---

Post content here ...
```

Another option is to simply set to `summary_large_image` if an image is set. Thoughts?

---

To see it in action, [go here](https://cards-dev.twitter.com/validator) and run the validator against https://www.grouparoo.com/blog/testing-session-storage-selenium-node. Then run it against https://www-grouparoo-com-git-twitter-card-type.grouparoo.vercel.app/blog/testing-session-storage-selenium-node.


evantahler (approved):
I don't really know anything about the types of twitter cards... so... 👍 !

PR #238, opened 2021-02-19
Homepage Messaging Revamp

evantahler:
Screen Shot 2021-02-18 at 6 17 42 PM

Screen Shot 2021-02-18 at 6 40 59 PM

Screen Shot 2021-02-18 at 6 22 08 PM

This PR updates the meta-information with the following:

```
const title = "Grouparoo: Open Source Data Synchronization Framework";
const description = "Grouparoo is an open source framework that helps you move data between your database and all of your cloud-based tools.";
const tagline = `Stop writing code to sync data to ${pluginName}*`;
```

The name of the destination plugin we pick on changes on each page load.

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,32 +1,37 @@
-import React from "react";
+import { useState, useEffect } from "react";
 import { Container, Button, Row, Col } from "react-bootstrap";
 import Image from "../components/Image";
 import Head from "next/head";
+import Link from "next/link";
 import HomepageCodeBox from "../components/home/homepageCodeBox";
 import IntegrationsSection from "../components/home/integrationsSection";
 import SmallIntegration from "../components/home/smallIntegration";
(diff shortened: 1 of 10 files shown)


pages/index.tsx
```diff
@@ -476,3 +506,8 @@ export default function IndexPage() {
     </>
   );
 }
+
+IndexPage.getInitialProps = () => {
+  const pluginName = randomPlugin("destination");
```
bleonard: I dig the randomness here, but I think we want to control it a bit more. Sailthru doesn't have as big on an impact and future ones will be more more niche.

Suggestion:
```js
export function homePagePlugin() {
  const names = [
    "Marketo",
    "Marketo",
    "Salesforce",
    "Salesforce",
    "Zendesk",
    "Mailchimp",
  ];
  return names[Math.floor(Math.random() * names.length)];
}
```

PR #276, opened 2021-03-02
Remove React warnings for tabs in docs

seancdavis:
So far I've only seen this as an issue where we've used tabs in MDX files. I submitted a PR to react-bootstrap (react-bootstrap/react-bootstrap#5686), but they suggested we build a local wrapper.

I wrote this wrapper as a one-off thing for now. It may pop up again in the future, at which point we could be a little smarter about this.


evantahler (approved):
Great work!

I guess that makes sense - MDX is storing extra properties to help with rendering while react-bootstrap is trying to tell you it doesn't know what those properties are...

PR #279, opened 2021-03-03
Add details for common spec helper methods

seancdavis:
Add documentation for commonly-used spec helper methods for Grouparoo users.

⚠️ This corresponds with grouparoo/grouparoo#1419 and contains a bug that won't be fixed until the next release. (Currently `truncate` does so at the wrong time and causes errors with code config projects.

Concerns/notes:

- I didn't automate this because I felt like it was important to be selective about the information shared with those who will consume this page. Maybe eventually we want to share the whole spec helper API, but for now it seems like we could stick to the common and simple examples.
- Some of the options seem to be geared toward usage in core alone. I was on the fence about removing the plugin options.
- I don't think I have great language for the options here.

Screen Shot 2021-03-03 at 3 02 22 PM

### DIFF
--- a/pages/docs/running/testing.mdx
+++ b/pages/docs/running/testing.mdx
@@ -101,3 +101,53 @@ describe("snapshot", () => {
 ## Testing Notes
 
 - Unless you specify in your `package.json` or `jest.config.js`, You'll need to use magic comments to let Jest know that we are running a Node.js app (`@jest-environment node`)
+
+## Spec Helper API
+
+The spec helper makes a few helpful methods available to you. Here are the commonly-used methods and options:
+
+### `grouparooTestServer`
+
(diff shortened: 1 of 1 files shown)


pages/docs/running/testing.mdx
```diff
@@ -101,3 +101,53 @@ describe("snapshot", () => {
...
+Starts a test version of the Grouparoo server prior to running the specs and stops it after running the tests.
+
+Usage:
+
+```js
+const options = {
+  truncate: false,
+  enableTestPlugin: false,
+  resetSettings: false,
+  disableTestPluginImport: false,
```
evantahler: I don't think that our customers will be using these options.  Maybe just don't list them?

evantahler:
@seancdavis V0.2.5 is out, so I merged this in!

PR #280, opened 2021-03-03
Integrations refactor

andyjih:


evantahler (changes requested):
YEAH!
Screen Shot 2021-03-04 at 3 12 16 PM

I think this is looking really good!  You also started to find some places to clean up (https://github.com/grouparoo/www.grouparoo.com/pull/280/commits/89cda2d14a75b5ed0297d76d78b16bcc9eeb819a).

I'll break up the big `data/inegrations.ts` into a file-per-plugin so it's easier to manage and find what you are looking for, but other than that, I think this is ready!  I'll break up the file and merge it!

evantahler (approved):

PR #298, opened 2021-03-10
Add sync as blog tag

evantahler:


evantahler:
@bleonard FYI - your new tag wasn't working in the category view

PR #300, opened 2021-03-10
Deployment Docs move to `app-example-*` repos

evantahler:
This PR removes the platform-specific deployment pages and instead directs people to our new `app-example-*` repos, which contain the same information and the example code.

Screen Shot 2021-03-10 at 1 41 27 PM


data/docs-nav.ts
```diff
@@ -57,9 +57,31 @@ const DocsNav: NavItem[] = [
...
+      {
+        title: "GCP",
+        path: "https://github.com/grouparoo/app-example-gcp",
+        target: "_blank",
+      },
+      {
+        title: "Docker",
+        path: "https://github.com/grouparoo/app-example-docker",
+        target: "_blank",
+      },
```
evantahler, line 84: @seancdavis well that was simple!

PR #306, opened 2021-03-14
Trying node v15

bleonard:
Trying to use v15 locally (Mac M1!), but Vercel does not support it yet.
The tests and github actions and prod/previews still run 14, so I want to see if this works.

It seems to be ok.


bleonard:
Screen Shot 2021-03-14 at 3 57 28 PM

bleonard:
Also

Screen Shot 2021-03-14 at 4 00 53 PM

bleonard:
Vercel trying to make lockfile v2 work.
Seems to be ok.

Screen Shot 2021-03-14 at 4 06 06 PM

evantahler:
I rebased from `main` and removed the `engines` section from `package.json` - since we set the node version explicitly with vercel, and this package isn't shared,  it's not needed there

PR #330, opened 2021-03-19
Adds /integrations/... to sitemap

bleonard:
* /integrations

other children to consider
* /jobs?
* /blog author
* /blog page 2+
* /whats-new page 2+
* /blog category


bleonard:
I've seen another approach where we actually build the whole app and then see what files exist. Unclear if that is better or not. If this approach gets out of hand, it's worth considering. For now, I feel ok because it's using the same `getStaticPaths`.

PR #331, opened 2021-03-19
Dogfooding blogpost

andyjih:


bleonard (approved):
Made the twitter card work.
I also remove "notes" tag, which has been "random things encountered while developing"

Note, you aren't actually showing the image.
You'd need to add an <Image> there. We've been putting `source=` on there and will figure out the right way to attribute it.

PR #343, opened 2021-03-25
[Blog Post] Three Methods for Promisifying Node Callback Functions

seancdavis:
I went deep on SQLite! After brainstorming for a bit, I think I actually have a (loose) series of four posts I can write all around the same topic.

Because I spent so much time on research, I wanted to start with the simplest of the four. There is some competition in keywords, but it's one of those listicle-type tittles that tends to play well.

👀  [Preview](https://www-grouparoo-com-git-177465446-promisifying-grouparoo.vercel.app/blog/promisifying-node-functions)


pages/blog/promisifying-node-functions.mdx
```diff
@@ -0,0 +1,102 @@
...
+const result = await doThingAsync("theThing");
+```
+
+That's super simple! And that's why this is my preferred approach when it can be implemented cleanly.
+
+When we're dealing with an instance of a class or object, we have to bind that object to `promisify`. Here's an example:
+
+```js
+const instance = new Thing();
+const doThingAsync = promifiy(instance.doThing).bind(instance);
```
pedroslopez: Typo!
```suggestion
const doThingAsync = promisify(instance.doThing).bind(instance);
```

evantahler:
Looks good, and I love the image!

There's one note about `util.promisify` that's bitten me in the past - the function you are promiseifying (that's a word)... really does need to be of the pattern `doThing('theThing', function(error, result)` - if `error` isn't the first argument to the callback things break.

evantahler (commented):
Looks good, and I love the image!

There's one note about `util.promisify` that's bitten me in the past - the function you are promiseifying (that's a word)... really does need to be of the pattern `doThing('theThing', function(error, result)` - if `error` isn't the first argument to the callback things break.

pages/blog/promisifying-node-functions.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+The old way looked something like this:
+
+```js
+doThing('theThing', function(error, result) {
+  // Catch the error or do something with result ...
+}
+```
+
+This pattern isn't compatible with the new approach where I want to wait for each asynchronously function to resolve so I can predict the order in which my code is executed.
```
andyjih:
```suggestion
This pattern isn't compatible with the new approach where I want to wait for each asynchronous function to resolve so I can predict the order in which my code is executed.
```

andyjih (approved):
looks good!

seancdavis:
Thank you, @andyjih @pedroslopez and @evantahler!

@evantahler That's a good point. I did notice that but I think it'd be good to call out as well. I'll add a note.

PR #384, opened 2021-04-15
Add tests to check for links not in sitemap

seancdavis:
This adds individual tests for every URL in the sitemap to ensure that all local links on the page point to a URL also included in the sitemap.

It also:

- Provides a way to allow patterns of URLs that may not be in the sitemap
- Adds blog authors and categories to the sitemap
- Adjusts canonical URL values on author and category pages
- Fixes name of Mongo what's new post


bleonard (commented):
This test is already doing through all the links on the site.
https://github.com/grouparoo/www.grouparoo.com/blob/main/__tests__/integration/html.ts#L89-L108

I was thinking that we could see if they are links to our site and basically make sure they are in the sitemap. 

Is that better than this approach?
This approach handles external links, but that wasn't my priority per se.

In either case, this could be a test instead of a github action.

seancdavis:
@bleonard I added a simple integration test to show the difference. Here are my thoughts after playing with both solutions ...

What I like about your approach:

- We already have a pattern for it.
- Doesn't require another build step.
- It's a relatively quick fix. Doesn't add much code.

Things I liked about my approach / challenges with your approach:

- I wasn't testing external URLs, but they _could_ be tested.
- We should probably follow hashes at the end of URLs to ensure there is an element with that `id` attribute on the page. To do this with your approach is certainly possible, but would require breaking up making all the requests with running the tests so that we could preload all available anchors for each path.
- My approach was making requests against the development server. This could be easily adjusted to running the server against the built project. Either way, it was looking for redirects and following them. This was intentional, so that we could keep routes like `/docs/config/properties` around, knowing that route may exist again one day.
- I can't quite tell from looking, but I'm not sure if your approach is covering SSR pages — i.e. if the sitemap includes SSR pages.
- My approach looks recursively through the site by following links. This is good and bad. It's good because it could catch pages not listed in the sitemap. It's not great because it could miss pages listed in the sitemap. We probably want to use a two-pronged approach here and start with the sitemap, but collect any stragglers we may have missed.

---

In other words, first we should answer these questions:

1. Is the sitemap a good enough representation of every page on the site that we want to test? Should we also follow links recursively throughout the site? (Bonus: If we find something not in the sitemap, should we fail the test suite on the grounds that the sitemap didn't have every page that we're linking to internally?)
2. Do we want to include tests for anchors at the end of paths?
3. Do we want to validate external links?
4. Are redirected paths that ultimately lead to a `200` response valid paths?

When we know the answer to these questions, then we can decide what the path of least resistance is. In either case, I agree that we should keep it within this test suite and the tools we already have.

seancdavis:
FWIW, my opinions are:

1. Use sitemap as a starting point, but recursively follow any internal links we find. Fail the test suite if we find an internal link that we're not including in the sitemap. We may one day want to be more nuanced in being able to ignore pages from the sitemap that we want to link to internally, but I don't see the benefit of that today. I see more benefit in ensuring all paths are indexable in the sitemap.
2. Yes.
3. Not right now. That would add quite a bit of overhead to the test suite, when the real value is making sure our internal links are valid.
4. I'm on the fence about this one. On one hand, they are valid links. On the other hand, I kind of like being super opinionated and saying if you're going to put a link on a page it has to return a `200` response.

seancdavis:
From a conversation with @bleonard we decided that we're going to:

- Use the test suite
- Ignore anchors in paths
- Use the sitemap as a source of truth
- Not crawl internal links
- Ignore external URLs
- Consider redirected paths invalid

seancdavis:
I have the tests working how I want. This has led to finding a series of URLs that are not captured in our sitemap. I'm going to balance manually allowing these paths with adding them to the sitemap. A bit more work to do here.

seancdavis:
Alrighty. This thing is ready for review again.

public/sitemap.xml
```diff
@@ -339,6 +339,36 @@
     >daily</changefreq><priority>0.6</priority></url><url><loc
     >https://www.grouparoo.com/integrations/destinations/zendesk</loc><changefreq
     >daily</changefreq><priority>0.6</priority></url><url><loc
+    >https://www.grouparoo.com/blog/author/evan_tahler</loc><changefreq
```
bleonard: In this PR, we decided that these author and tag pages were duplicative and left them out of the sitemap. https://github.com/grouparoo/www.grouparoo.com/pull/330

maybe they are in the whitelist or otherwise ignore in the test.

seancdavis: Okay. That's fine. I kept the changes I made in terms of moving around the source of where these routes are generated, so it'd be trivial to add them to the sitemap in the future, should they become more unique.

PR #399, opened 2021-04-20
added comp page for Census

teallarson:
I added a few comments [here](https://docs.google.com/document/d/1JAjJgcq3ohcIX3ZxrmxbAXUgRsyQfCLFN6ZBT4GW08g/edit#) that I think would clarify a few points in the copy.  

This PR has the copy as-is.


pages/solutions/census-alternative.tsx
```diff
@@ -0,0 +1,148 @@
...
+              <h3>💰 Save bags of money</h3>
+              <p>
+                Grouparoo is open source and free to run. Stop being charged by
+                event volume and focus on what you want to build.
+              </p>
+              <br />
+            </Col>
+          </Row>
+        </Container>
+      </div>
```
seancdavis, line 124: How similar is this section to the Segment page? Is there an opportunity to share a component? Or maybe create a component and pass the content in as props? Then we don't have to duplicate the markup on every comparison page.

pages/solutions/census-alternative.tsx
```diff
@@ -0,0 +1,148 @@
+import React from "react";
+import { Container, Button, Row, Col } from "react-bootstrap";
+import Head from "next/head";
+import GetStarted from "../../components/home/getStarted";
+import hydrate from "next-mdx-remote/hydrate";
+import { loadMdxFile } from "../../utils/mdxUtils";
+
+const components = {};
+
+export default function VsCensus({ pageProps }) {
```
seancdavis, line 10: I'm wondering how similar this is with the Segment page? I mentioned sharing a section below, but now I'm wondering if this should be a templated approach?

IOW building a template at `pages/solutions/[solution].tsx` and then using individual MDX files for each solution. And we can use frontmatter for some of more unique sections.

Maybe that's overkill at this point, but my gut is (without looking) that a lot of the code was duplicated.

seancdavis (approved):
Overall, it looks good to me. I'd be fine merging as is. I left a couple comments on how you might think about DRYing up the code. Happy to explore with you if you have questions or what to bounce ideas around.

pages/solutions/census-alternative.tsx
teallarson: I like that idea better.  I'd built out separate components for the reusable parts, but reverted that commit because it should really be a separate PR.  A template sounds like a tidier solution overall.  I'll put that together and then start working on visual elements (checkboxes, table, etc.)

pages/solutions/census-alternative.tsx
teallarson: Yep, I had done that but thought it deserved a separate PR.  I like the layout idea below as well.

teallarson:
See https://github.com/grouparoo/www.grouparoo.com/pull/402 instead.  Went with dynamic rendering per @seancdavis suggestion.

PR #402, opened 2021-04-20
Comparison pages

teallarson:
This PR replaces [PR #399](https://github.com/grouparoo/www.grouparoo.com/pull/399).

Comparison pages can now be rendered dynamically from .mdx.  We may want to adjust some (the header copy is very similar among pages currently), but working for now!

Also added tests for the comparison pages we have so far (Census and Segment)


__tests__/integration/comparisons.ts
```diff
@@ -0,0 +1,44 @@
...
+     const env = await helper.prepareForIntegrationTest();
+     url = env.url;
+   }, 1000 * 60);
+ 
+   afterAll(async () => {
+     await helper.shutdown();
+   });
+
+ 
+   test.each(cases)("it renders the comparison page for %p", async (pageSlug) => {
```
evantahler, line 30: Nice dynamic test!

components/footer.tsx
```diff
@@ -117,9 +117,17 @@ export default function Footer() {
                   <a>Healthcare</a>
                 </Link>
                 <br />
-                <Link href="/solutions/segment-alternative">
+                <Link href="/solutions/segment">
```
evantahler, line 120: @andyjih do you think that having the `alternative-` in the URL is important?

pages/solutions/[id].tsx
```diff
@@ -3,24 +3,34 @@ import { Container, Button, Row, Col } from "react-bootstrap";
...
+  const router = useRouter();
+  const comp: ComparisonInfo = pageProps.comparison;
+  const content = hydrate(comp.source, { components });
+  let possessive = "";
+
+  if (comp.competitor.charAt(comp.competitor.length - 1) == "s") {
+    possessive = "'";
+  } else {
+    possessive = "'s";
+  }
```
Comment (evantahler) [inline pages/solutions/[id].tsx]: Excellent!  I would have certainly missed this.

public/sitemap.xml
```diff
@@ -149,13 +149,17 @@
     >daily</changefreq><priority>0.6</priority></url><url><loc
     >https://www.grouparoo.com/integrations/postgres-to-salesforce</loc><changefreq
     >daily</changefreq><priority>0.6</priority></url><url><loc
+    >https://www.grouparoo.com/solutions/census</loc><changefreq
+    >daily</changefreq><priority>1.0</priority></url><url><loc
```
evantahler, line 153: Excellent! The sitemap found the new pages

utils/comparisonPages.ts
```diff
@@ -0,0 +1,32 @@
...
+    pro2: string;
+    pro3: string;
+    pro4: string;
+    comp1: string;
+    comp2: string;
+    source: any;
+}
+
+
+export async function getComparison(slugId): Promise<ComparisonInfo> {
```
evantahler, line 13: Good use of typescript types in the function definition

evantahler (approved):
Overall, this looks great!

One thing that I'm curious about is the `pro1, pro2, pro3` in the frontmatter, vs an array of `pros`.  Would that be easier to work with?  Or, are we sure that every page has exactly 3 pros?  Either way, that can be addressed in the next story.

teallarson:
> Overall, this looks great!
> 
> One thing that I'm curious about is the `pro1, pro2, pro3` in the frontmatter, vs an array of `pros`. Would that be easier to work with? Or, are we sure that every page has exactly 3 pros? Either way, that can be addressed in the next story.

From a writing/parallel structure perspective I was thinking three each... but an array sounds more realistic and useful overall.

andyjih (approved):
Awesome work! 👍🏽

PR #409, opened 2021-04-28
Updating UI on comparison pages

teallarson:


teallarson:
circleCI failure is because circleCI doesn't like my test.todos

teallarson:
TODO:  
* taking a second look at testing this afternoon with @evantahler  (I think it could be refactored, though it does work in its current state)
* fix spacing on cells with two items (eg: with checkmark AND text)
* fix section dividers on 1920x1080 + screens

teallarson:
Will push as-is for now.  

TODO's:  
- I don't like the current look of boxes with a ✅ AND text in the table
- weird banner appearance on 1920x1080+ screens
- would like to make the test suite for these pages more efficient.  
 
BUT all is running, so I'm thinking it can go up as is and be improved as we go.

components/comparisons/comparisonRow.tsx
```diff
@@ -0,0 +1,66 @@
+import Image from "../Image";
+
+export default function ComparisonRow({ rowData }) {
+  // rowData = {
+  // feature: featureName,
+  // grouparoo: true,
+  // comp: false
+  // }
```
evantahler, line 8: Maybe remove these comments

components/comparisons/comparisonTable.tsx
```diff
@@ -0,0 +1,38 @@
...
+import ComparisonRow from "./comparisonRow";
+
+export default function ComparisonTable({ comparisonData }) {
+  // comparisonData = [{
+  // category: [{
+  //   feature: featureName,
+  //   grouparoo: true,
+  //   comp: false
+  // },]
+  // }]
```
evantahler, line 12: If you did want to keep this information around, you could make a typescript type, and then say that the params you were getting into the component were of that type:

```ts
import {ComparisonData} from '../utils/comparisonPages'

export default function ComparisonTable({ comparisonData } : {comparisonData: ComparisonData}) {
  // ...
}

```

pages/solutions/[id].tsx
```diff
@@ -2,12 +2,15 @@ import React from "react";
...
+import FeatureBanner from "../../components/comparisons/featureBanner";
 import hydrate from "next-mdx-remote/hydrate";
 import { useRouter } from "next/router";
+import { PageNavigation } from "../../components/pageNavigation";
 import { ComparisonInfo, getComparison } from "../../utils/comparisonPages";
 import { arrayToReadableList, possessiveNoun } from "../../utils/inflectors";
-
-const components = {};
+import * as components from "../../components/comparisons";
+// import { TableOfContents } from "../../components/docs/tableOfContents";
```
Comment (evantahler) [inline pages/solutions/[id].tsx:14]: More un-used comments

evantahler (approved):
Looks good!

Some comments below for cleanup and a suggestion that you could type some of your inputs for clarity if you wanted.

andyjih (approved):
👍

components/comparisons/comparisonTable.tsx
tealjulia: As a workflow question.... I agree these changes should be made, but you and Andy approved it as is... do I make the changes then squash and merge?  OR just squash and merge and tidy those up in the next iteration (low-stakes here as it's just removing unneeded comments...)

PR #417, opened 2021-04-30
Updated Nav Bar and Fixed Banners

teallarson:
- added multi-column dropdown called "Solutions" in place of integrations
  - matched top-level nav styling
  - custom two-column dropdown built/styled
  - links organized by solution category


teallarson:
This PR also includes fixing the section divider banners to fit full width past 1400px

pages/solutions/[id].tsx
```diff
@@ -82,18 +80,12 @@ export default function Comparison({ pageProps }) {
         </Container>
       </div>
       <FeatureBanner features={comp.features} />
+      <Container className="col-9 col-md-4"></Container>
```
Comment (evantahler) [inline pages/solutions/[id].tsx]: I think this empty container is doing nothing?

evantahler (changes requested):
Notes:
* The Dropdown has a good layout on mobile too 
* Aesthetically, perhaps the headers in the dropdown need either increased padding/boldness.  On mobile (attached), it seems like they are very close to the same font as the rest of the menu (Attached)

Bugs:
* Solutions/redshift doesn't go to the right page
* Looks like there are some git troubles here.  This PR has commits about adding comparison pages too. 

IMG_785434C4AE74-1

pages/solutions/[id].tsx
Comment (teallarson) [reply pages/solutions/[id].tsx]: Ah, thanks.  Stray container from trying to add in-page navigation.

Comment (teallarson) [reply pages/solutions/[id].tsx]: > Notes:
> 
> * The Dropdown has a good layout on mobile too
> * Aesthetically, perhaps the headers in the dropdown need either increased padding/boldness.  On mobile (attached), it seems like they are very close to the same font as the rest of the menu (Attached)
> 
> Bugs:
> 
> * Solutions/redshift doesn't go to the right page
> * Looks like there are some git troubles here.  This PR has commits about adding comparison pages too.
> 
> IMG_785434C4AE74-1

Yes, I branched off of a branch so the commit history is jumbled.  What would be the right way to fix that without losing changes?  A rebase?

__tests__/integration/comparisons.ts
```diff
@@ -26,56 +26,61 @@
...
-    const H3s = await browser.findElements(by.tagName("h3"))
-    const H3Texts = await Promise.all(H3s.map(element => element.getText()));
-    expect(H3Texts).toContain(`Easily define profiles, no matter your data.`)
-
-  });
+async function sleep (time=1000) {
+  return new Promise((resolve) => {
+    setTimeout(() => resolve(null), time)
+  })
+}
```
evantahler, line 83: @evantahler and @tealjulia looked into making this test better, but we:
* Could not get the markdown to render in component tests, so we had to stick with integration tests
* Found a wacky problem with selenium that needed a sleep() to fully render the page.

There's more work to do here to make these better, but we can solve that later and get these updates out

evantahler (approved):
Screen Shot 2021-05-03 at 6 35 57 PM

Looks good!

teallarson:
Merging per discussion with @evantahler

PR #428, opened 2021-05-06
178045142 - Adjust top level nav to include integrations

teallarson:
TO DO:
* After category pgs are approved, add to "solutions" dropdown

DONE:
* Add nav-link class to navigation links
* Add top level Integrations link
* Remove integrations from solutions menu
* Adjust vertical alignment


evantahler (changes requested):
Good start!

1. [to-check] I think we wanted the Github button to take folks to the github repo, in an attempt to prove we are open source and ask them to star it.  I defer to @andyjih if that's still the plan
2. [idea] With the wider spacing between links, it /almost/ looks like the nav is center-aligned, but not quite.  If the intent was to add more padding, perhaps center aligning might help 
3. [bug] On the "tablet" size screen, a few things are wonky (attached):
   1. The Github button goes to 2-lines
   2. The "get started" button is out-of-frame
   3. You may need to hide an element to make everything fit.  We do this now, dropping "community" when the screen gets too small, but we aren't yet on mobile to get the hamburger.
4. [bug] The hover states are a bit off-center (attached)
Screen Shot 2021-05-06 at 7 39 11 PM

Screen Shot 2021-05-06 at 7 39 48 PM

andyjih:
> 1. [to-check] I think we wanted the Github button to take folks to the github repo, in an attempt to prove we are open source and ask them to star it.  I defer to @andyjih if that's still the plan

Correct-- the link should continue to go to the github repo please! The styling also changed-- the goal was to make it look as close to the Github star button, so I'd like that to stay the same. 

Also, the nav feels a bit crowded now, so an idea is to remove the link to the blog from the top-nav. We should keep it in the footer though.  

Looks good!

components/navigation.tsx
```diff
@@ -31,189 +32,231 @@ export default function Navigation() {
...
+                      </Dropdown.Item>
+                      <Dropdown.Item>
+                        <Link href="/integrations/destinations/marketo">
+                          <a className="nav-link" role="button">
+                            Marketo
+                          </a>
+                        </Link>
+                      </Dropdown.Item>
+                      <Dropdown.Item>
+                        <Link href="/integrations/destinations/zendexk">
```
evantahler: Typo!

evantahler (changes requested):
Other than that broken link, 👍

evantahler (approved):

evantahler (changes requested):
There's something not quite right with git again - there are too many commits/changes here.  For example, the layouts for the Comparison and Category pages have already been merged in, along with the blog post.

teallarson:
Worked with @seancdavis to fix commit history and we ended up switching to a new branch/PR #433

PR #430, opened 2021-05-08
Adjusted Pricing Page UI

teallarson:
On pricing page, in mobile view, the features table was overflowing the width of the frame and causing screen-width components to appear less than screen-width (eg: footer was truncated).

Solved by removing "Hosted" column on mobile.  Its contents are the same as "Enterprise", so I believe the page is fine without on mobile.

UI remains the same for tablet and up.

### DIFF
--- a/pages/pricing.tsx
+++ b/pages/pricing.tsx
@@ -149,14 +149,14 @@ export default function PricingPage() {
       <div id="featureComparisons" className="featureTableSection">
         <Container>
           <Row>
-            <Col>
-              <Table style={{ textAlign: "center" }}>
+            <Col className="mw-95vw">
+              <Table className="text-center w-auto">
                 <thead>
                   <tr>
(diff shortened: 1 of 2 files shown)


evantahler (changes requested):
There are some shenanigans going on here:

1. at the 'desktop' breakpoint the table is off-center and perhaps to narrow

Screen Shot 2021-05-07 at 6 54 52 PM

2. On the Tablet? size... these boxes should probably start to stack 

Screen Shot 2021-05-07 at 6 54 45 PM

teallarson:
Got it!  I think React Bootstrap is based on Bootstrap 4 still, but Bootstrap dropped Card Decks in v5 in favor of their grid classes -- they're a lot more adjustable for something like this so that's what I've switched us to!

evantahler (approved):

PR #433, opened 2021-05-11
Adjusted top level nav

teallarson:
This replaces PR #428.  @seancdavis and I were unable to get the merge/commit history tidied as nicely as we wanted. 
 Concluded it was simpler just to move the changes to a new branch.  In this commit:

* Adjusted top level nav to include both "integration" and "solutions" mega-menus (two column dropdowns) on tablet/dekstop and accordion-style submenus on mobile.
* Github button toggles icon/no icon and visible/hidden depending on screen size

### DIFF
--- a/components/navigation.tsx
+++ b/components/navigation.tsx
@@ -261,7 +261,7 @@ export default function Navigation() {
               </Nav.Link>
             </Nav>
             <Button
-              variant="outline-light"
+              variant="outline-dark"
               href="https://github.com/grouparoo/grouparoo"
               size="sm"
               className="col-sm-12 d-md-none d-lg-inline-block col-lg-1 mx-lg-1 mx-0 mt-3"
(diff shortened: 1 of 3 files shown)


components/navigation.tsx
```diff
@@ -31,189 +32,255 @@ export default function Navigation() {
...
+                  </a>
+                </Link>
+              </Nav.Link>
             </Nav>
+            <Button
+              variant="outline-dark"
+              href="https://github.com/grouparoo/grouparoo"
+              size="sm"
+              className="col-sm-12 d-md-none d-lg-inline-block col-lg-1 mx-lg-1 mx-0 mt-3"
+            >
```
evantahler: I believe we still want this link to open in a new tab (`target="_blank"`)

evantahler (changes requested):
Much better!

Just a minor change to keep the Github link opening in a new tab, and this is good to go!

components/navigation.tsx
teallarson: Thanks for catching that one!  Updated.

Screen Shot 2021-05-11 at 11 01 48 AM

evantahler (approved):

PR #447, opened 2021-05-14
[blog] - varchar 191 post

evantahler:
Fun Facts: 
* Markdown has footnotes!?
* I took that library photo - I give me the rights to use it!


pages/blog/varchar-191.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+```sql
+select id from users where email = 'foo@example.com';
+```
+
+As your table gets bigger, searches get slower because your database has to check every row to find a match. However, if you add a **search index**, you are telling your database to essentially "pre-compute" popular search patterns with a tree so the next search is much faster. In essence, indexes spend computation time (and a little bit of disk space) making writes to the database slower, to speed up reads later. For most applications this is a great tradeoff, since they are "read heavy" and "write lite".
+
+So, why use `varchar`? Indexes can be made to preform better when assumptions can be made about the type of data they store. Knowing how long the strings in the index are is one of the best ways speed things up. For some databases, you aren’t allowed to add a search index to columns of type `text` becaus this optimization can't be done, while in others, the index just won’t preform as well. In fact, historically, databases were constructed with limits on how big an index could be to optimize search and how they stored data on disk.
```
andyjih:
```suggestion
So, why use `varchar`? Indexes can be made to perform better when assumptions can be made about the type of data they store. Knowing how long the strings in the index are is one of the best ways speed things up. For some databases, you aren’t allowed to add a search index to columns of type `text` because this optimization can't be done, while in others, the index just won’t perform as well. In fact, historically, databases were constructed with limits on how big an index could be to optimize search and how they stored data on disk.
```

pages/blog/varchar-191.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+## It’s MySQL’s fault
+
+Ok, so indexes are good. But, generically, it seems that an index of _any_ size should work, and while that’s true today, it wasn't always possible. The next stop on our journey is to look at what the default column size was far in the past, and that was 255 characters, e.g.:
+
+```sql
+email_address varchar(255) NOT NULL
+```
+
+MySQL, the most popular open source database of the early 2000’s had a limit of 255 characters in indexed fields. The history is fuzzy as to why MySQL chose a 255 character limit (see the articles linked below), but the most popular theories include:
```
andyjih:
```suggestion
MySQL, the most popular open source database of the early 2000s had a limit of 255 characters in indexed fields. The history is fuzzy as to why MySQL chose a 255 character limit (see the articles linked below), but the most popular theories include:
```

pages/blog/varchar-191.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+## It’s 🐟’s fault
+
+255 makes a lot more sense than 191. How did we get to 191? I’m going to blame emoji 😜. Seriously. Well, `utf8mb4` at least, the charter set that allows for "international" [^1] characters, and included the first emoji. MySQL in the early 2000’s was happy supporting 255 characters in `varchar` columns and indexing them. However, the the most popular text encoding (`Latin1` or `utf8` ) on the most popular MySQL database engine (`innodb`) assumed that 3 bytes was enough to store every character [^2], and once `utf8mb4` came along with characters like 𠼭 and 🐟, 4 bytes were needed to store each character. There were [more character to choose from](https://www.fileformat.info/info/charset/UTF-8/list.htm), so referencing them took more bytes.
```
andyjih:
```suggestion
255 makes a lot more sense than 191. How did we get to 191? I’m going to blame emoji 😜. Seriously. Well, `utf8mb4` at least, the charter set that allows for "international" [^1] characters, and included the first emoji. MySQL in the early 2000s was happy supporting 255 characters in `varchar` columns and indexing them. However, the the most popular text encoding (`Latin1` or `utf8` ) on the most popular MySQL database engine (`innodb`) assumed that 3 bytes was enough to store every character [^2], and once `utf8mb4` came along with characters like 𠼭 and 🐟, 4 bytes were needed to store each character. There were [more character to choose from](https://www.fileformat.info/info/charset/UTF-8/list.htm), so referencing them took more bytes.
```

pages/blog/varchar-191.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+The way `innodb` MySQL databases worked was that you can only have 767 bytes for an index - enough to store 255 3-byte characters (`767/3 = 255`). This is an extreme example of index optimization where knowing the size of the data you are indexing matters! So if the characters took more space to store, then the number of characters you could index had to get smaller. Specifically, `767/4 = 191` characters! As more software supported an international audience, `varchar(191)` replaced `varchar(255)` as the default.
+
+[^1]: "International" is an odd way to talk about the languages that most of the world uses, specifically those using non-latin asian characters. However, operating systems and databases had thoroughly english-centric origins, and the legacy of those early choices is still rippling though our code today.
```
andyjih:
```suggestion
[^1]: "International" is an odd way to talk about the languages that most of the world uses, specifically those using non-Latin characters. However, operating systems and databases had thoroughly english-centric origins, and the legacy of those early choices is still rippling though our code today.
```

andyjih (approved):
Looks good! Made some minor edits.

PR #464, opened 2021-05-20
Integrations update

teallarson:
- Added SQLite, MongoDB, Pardot cards to integrations page
- Adjusted filter so NextJS doesn't generate empty pages for them in the background
- Adjusted and added testing


teallarson:
A few notes: 
- sqlite page already had quite a bit of information in it, but it was commented out.  I've preserved that.
- the reason for adding empty paragraphOne and paragraphTwo has to do with how the PluginData filter determines who to build cards (and pages) for.  I suppose there are other ways to adjust these.  This way works so we can have (1) plugins with no page and no card by leaving the object out (ie: csv) (2) plugins with a card and no page by putting an object with empty strings (ie: pardot) (3) plugins with a card and a page by completing the object (and adding the additional data needed)

evantahler (commented):
Looks good - the new plugins are on the page!

A few questions:
1. Should there be a visual distinction that some plugins don't have a page?  Maybe the box should be greyed out or something?
2. Using the existence of the paragraphs to determine which plugin gets a page seems unintuitive - you could add more properties to each plugin type, eg: `hasIntegrationPage: false` and then they all show up in the grid

bleonard (approved):
Looking good. Just csv and google sheets as you noted

teallarson:
> Looks good - the new plugins are on the page!
> 
> A few questions:
> 
> 1. Should there be a visual distinction that some plugins don't have a page?  Maybe the box should be greyed out or something?
> 2. Using the existence of the paragraphs to determine which plugin gets a page seems unintuitive - you could add more properties to each plugin type, eg: `hasIntegrationPage: false` and then they all show up in the grid

1.  We could do that.  Currently, it just doesn't display as a link (so if you hover, there is no hand cursor and no "pop" off the page).  Could make it more obvious.

2. It turns out there is no reason we would want a plugin to be there with no card displaying.  That made the filtering tidier.  Now if there are the properties it needs for a page, it builds one, otherwise, it just builds the card now.

3. On a similar note, there are cards for CSV and Google Sheets now

evantahler:
> > Looks good - the new plugins are on the page!
> > A few questions:
> > 
> > 1. Should there be a visual distinction that some plugins don't have a page?  Maybe the box should be greyed out or something?
> > 2. Using the existence of the paragraphs to determine which plugin gets a page seems unintuitive - you could add more properties to each plugin type, eg: `hasIntegrationPage: false` and then they all show up in the grid
> 
> 1. We could do that.  Currently, it just doesn't display as a link (so if you hover, there is no hand cursor and no "pop" off the page).  Could make it more obvious.
> 2. It turns out there is no reason we would want a plugin to be there with no card displaying.  That made the filtering tidier.  Now if there are the properties it needs for a page, it builds one, otherwise, it just builds the card now.
> 3. On a similar note, there are cards for CSV and Google Sheets now

👍 makes sense, and good enough for now!

evantahler (approved):

PR #468, opened 2021-05-21
[Blog Post] We've gone too far into the frontend

pedroslopez:


pages/blog/weve-gone-too-far-into-the-frontend.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+This may seem simple at first, but it means that doing a simple action like changing the group, a number of things need to happen. If the tracked group has changed, we may need to kick off a run to get the latest data for all the affected profiles and export them to the destination. Similarly, if the destination is no longer tracking a group, we need to go to the destination and remove all the profiles that had been sent over in previous exports.
+
+To support this, our API had exposed three endpoints:
+
+- One for updating the destination's data, setting options and configuring which properties would be sent over.
+- Another for setting the newly-tracked group and kicking off an import/export of profiles, if needed.
+- Another for removing the previously-tracked group from the destination and kicking off an export to remove the profiles for the destination, if needed.
+
+Our simple button at the bottom of the page to save destination data now had to determine if the group had been changed and execute a number of API requests.
```
evantahler: You can't see the button in the screen shot - maybe fake it in Figma or mess with the HTML?

pages/blog/weve-gone-too-far-into-the-frontend.mdx
```diff
@@ -0,0 +1,84 @@
+---
+title: "We've gone too far into the frontend"
```
evantahler: Ha! This title is a bit sensational (which is maybe what you wanted).  A more accurate title I think is your H2 below - "The frontend is doing too much".  Maybe... "Your smart fontend is too smart" or  "The frontend knows nothing about state"?   Or... keep it sensational for the clicks :D

pages/blog/weve-gone-too-far-into-the-frontend.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+**↔️ State changes between requests**
+
+Another problem that comes with issuing multiple requests is that there's no guarantee that the state will be the same when that second request arrives to the backend. What if another user issued a request to update the same resource and it happened to be processed before yours? The state of that resource could be different at this point, meaning that your second request could produce unintended results.
+
+A very similar issue was happening at Grouparoo, where the group that the client _thought_ the destination was tracking was actually some other group. This meant that we could potentially miss triggering an export in certain scenarios.
+
+## The frontend is doing too much
+
+So how do we solve this problem? Ultimately, the solution lies in good API design. Though there are things that can and should be controlled in the frontend, things like these should really be decided directly in the backend. With this comes a balancing act in API design to make sure the clients have the right tools to accomplish what they need to do, while still supporting various use cases without affecting long-term maintainability.
```
evantahler: > things like these should really be decided directly in the backend.

Which things?  What "rule of thumb" could I be using to make those choices?  
* Things in which state changes between request?  
* Models that aren't "locked" to my session?
* Things that are slow?
* ... you get the idea

pages/blog/weve-gone-too-far-into-the-frontend.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+### Taking it back to RPC
+
+In the days of [RPC](https://en.wikipedia.org/wiki/Remote_procedure_call), clients would call functions on the backend that did exactly what they needed to do. This resulted in a somewhat "client-first" approach, where these functions would be tailor-made to support the client's functionality. For example, in this case you could imagine having an endpoint `/updateDestinationAndTrackGroup` that takes inputs as-is, which means we can have all these things correctly decided on the backend through a single request.
+
+Obviously this _also_ means that the client and server end up being tightly coupled, which introduces a whole other set of issues. In addition, it can be easy to end up with tons of similar functions to support slightly different use cases, which can be hard to maintain.
```
evantahler: ... and you would have a ton of API routes

evantahler (commented):
I think you have the bones of a good post, but as you said, the "thesis" needs to be made a little more clearly.  Here's my interpretation of your argument, and a shuffling of the order of the supports.  Also, I may have misinterpreted everything, and if I did, let me know! 

---

Thesis: Don't be a REST purist / Smart clients are too smart. 

1. There's a spectrum when designing APIs between exposing "methods" (RPC) and exposing "objects" (REST) and letting the clients do whatever they want.  The trend these days is more object-driven (REST), and that works out well with "smart" clients.  But... there are times when you can get in trouble.   
2. It's hard to tell where on that spectrum you should be.  List of things that can go wrong (network / partial failures / state) that may make you want to lean more to RPC for some things.
3. Here's a grouparoo story about how we leaned to far into REST - (Destination data page)
4. Here's what we did to fix it (tighter controls on client expectations;  no longer assuming state could be maintained by the client)
5. Fun fact - it's common for "real" products not to be fully RESTful anyway (Hubspot example)
6. Summary: Keep your clients dumber; good APIs allow clients to fully declare state vs the steps to get you get there.

pedroslopez:
@evantahler  Yeah, that's the gist of it! I restructured it a bit based on your suggestions to hopefully make things clearer. Thanks!

evantahler (approved):
Yes! Much clearer (and 100% more GIFs!).

PR #470, opened 2021-05-21
Added cta to main layout

teallarson:


seancdavis (commented):
It looks a little weird to me that the content partially overlaps the columns above, but only _barely_. It feels like it's floating.

I'd suggest either putting a background on this section to separate it from the content above, or keep it in the righthand column.

Screen Shot 2021-06-03 at 12 05 01 PM

teallarson:
> It looks a little weird to me that the content partially overlaps the columns above, but only _barely_. It feels like it's floating.
> 
> I'd suggest either putting a background on this section to separate it from the content above, or keep it in the righthand column.
> 
> Screen Shot 2021-06-03 at 12 05 01 PM

Yep.  That's what Evan and I landed on yesterday, too.

teallarson:
I landed on two potential solutions.  @seancdavis brought up the idea that since it's so widely used, perhaps it can be integrated as part of the page layout near the footer.  If we choose this route, I would think we would want the option to hide it on certain pages.  (There aren't many it's currently hidden on.  The docs landing page, the blog/blog posts, and /meet are the ones I could find)

Screen Shot 2021-06-03 at 10 58 25 AM

The other solution @evantahler and I and @seancdavis and I talked on was to keep it in-column.  This one is not my favorite personally:

Screen Shot 2021-06-03 at 11 07 15 AM

Thoughts? I think @andyjih has been working on our CTAs site-wide too.  Does this interfere or affect anything you're working on?

evantahler:
The top option (site-wide) makes a lot of sense to me!  This way, @andyjih  would only have one place to make a change if he wants to as well.

teallarson:
Most recent iteration uses the full-width CTA site-wide.  The layout takes an array of paths that will not display the CTA component.  currently it contains:   /blog, /docs, and /meet as those do not display the CTA on the page as of today.

seancdavis (commented):
I like the approach! (Not just because I suggested it, either. 😉 )

But it looks a little goofy on pages where there is a solid section above it, like the home page. There's a lingering border hanging out.

I'd suggest either a solid background that differs from the footer and is likely to differ from other pages and/or a `theme`/`variant` type of prop, where an individual page could set the style of the CTA from a set of options.

Screen Shot 2021-06-04 at 6 08 34 AM

teallarson:
Yep.  Here's the tweak on the most recent commit, going with a neutral, contrasting background across the board:

On a page with a light div above (like solutions):
Screen Shot 2021-06-04 at 8 24 26 AM

On a page with a dark div above (like homepage): 
Screen Shot 2021-06-04 at 8 24 41 AM


On a page with two-column layout above (like docs):
Screen Shot 2021-06-04 at 8 25 01 AM

evantahler (approved):
I like it!

seancdavis (approved):
Me too!

PR #472, opened 2021-05-25
Add reference doc for Snowflake plugin.

pauloouriques:


evantahler (changes requested):
This is a good start (and shout out to @seancdavis for making adding new pages easy)!

For snowflake, I think that I'd want the examples to showcase how all the columns and tables are CAPITAL_LETTERS.  Perhaps these shared components can provide an optional override?

evantahler (approved):

PR #481, opened 2021-05-28
adding examples for group rules to docs

teallarson:
- Added examples for all group rule types
- Adjusted layout to increase responsiveness
- Removed `build_static_refs` from core in [this PR](https://github.com/grouparoo/grouparoo/pull/1889) as we no longer want these .json files auto generated

BEFORE on mobile:  
Screen Shot 2021-06-02 at 10 38 06 AM

AFTER:
desktop/laptop/ipad pro
Screen Shot 2021-06-02 at 10 34 33 AM

mobile/ipad
Screen Shot 2021-06-02 at 10 47 25 AM

Some longer code lines (a few date rules) still require horizontal scrolling on narrow phone screens, but I'm hesitant to decrease the font size on mobile.  I think it's still an improvement.

Screen Shot 2021-06-02 at 10 47 09 AM

### DIFF
--- a/components/docs/ruleOpsTable.tsx
+++ b/components/docs/ruleOpsTable.tsx
@@ -6,6 +6,8 @@ import {
   Tabs,
   Tab,
   Table,
+  Row,
+  Col,
   NavItem,
 } from "react-bootstrap";
 import pgRuleOps from "../../data/property-ops-dictionary--postgres.json";
@@ -79,7 +81,7 @@ const DataTable = ({ data, heading, eventKey }) => {
                 format: <br />
                 <span className="pl-4">
                   <b>YYYY-MM-DDThh:mm:ss.mmm+TZD</b>{" "}
-                  ("2020-09-01T08:15:00+00:00")
+                  ("2020-09-01T08:15:00+00:00"){" "}
                 </span>
                 A date passed as YYYY-MM-DD will pass validation, but may result
                 in less accurate group results.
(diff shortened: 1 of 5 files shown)


components/docs/ruleOpsTable.tsx
```diff
@@ -64,14 +73,27 @@ const DataTable = ({ data, heading, eventKey }) => {
       <Accordion.Collapse eventKey={eventKey}>
         <Fragment>
           <p>
+            {heading === "date" ? (
+              <Alert variant="primary">
+                💡 Dates used for calculated groups should use full ISO-8601
+                format: <br />
+                <span className="pl-4">
+                  <b>YYYY-MM-DDThh:mm:ss.mmm+TZD</b>{" "}
+                  ("2020-09-01T08:15:00-8:00")
```
evantahler: I'd suggest using `+00:00` in your examples, keeping everything in UTC. Most databases are in UTC

evantahler (commented):
I think this is great!

I agree that everything probably needs an example. There's another story (https://www.pivotaltracker.com/story/show/177599093) for that, but please feel free to pick it up now as well!   Examples should be added to the table (like you are doing now), and also higher up on the page we have an example (`async function buildConfig`) with no rules.  Please add a group that shows off some of the more interesting rules, like "New High Value customers) - has an email address, isn't from @grouparoo.com, has signed up within 7 days, etc.

components/docs/ruleOpsTable.tsx
evantahler: But I think the warning you added about dates and your example format is 👍

teallarson: > I think this is great!
> 
> I agree that everything probably needs an example. There's another story (https://www.pivotaltracker.com/story/show/177599093) for that, but please feel free to pick it up now as well! Examples should be added to the table (like you are doing now), and also higher up on the page we have an example (`async function buildConfig`) with no rules. Please add a group that shows off some of the more interesting rules, like "New High Value customers) - has an email address, isn't from @grouparoo.com, has signed up within 7 days, etc.

Would you like me to add those additional examples and such to this PR or a new one?  I've been working through them Friday/this morning.

evantahler: Totally up to you - whichever is easier at this point.  It's related enough work that being in the same PR would be OK

data/property-ops-dictionary--postgres.json
```diff
@@ -2,322 +2,464 @@
   "boolean": [
     {
       "op": "exists",
-      "description": "exists with any value"
+      "description": "exists with any value",
+      "example": "{\n  \"propertyId\": \"subscribed\", \n  \"operation\": {\"op\": \"exists\" }\n}",
+      "caption": "All profiles that do not have data about subscription status"
```
evantahler: This file is actually generated from core, via https://github.com/grouparoo/grouparoo/blob/main/core/bin/build_static_refs.  The goal was to allow core to be the keeper of these rules, and then to have a way to extract and copy them in to this site.  However, we've learned that we need this extra information (example, caption) that core doesn't have. 

Does that mean that:
A. The example & caption should be moved into core so we can re-use it elsewhere
B. There's sufficiently extra/different information that the www site needs that core doesn't have so we should no longer build_static_refs

cc @seancdavis - what do you suggest?

evantahler (commented):
The UI and content changes are great.  However, we should discuss where this data should live a little further (see comment below)

data/property-ops-dictionary--postgres.json
seancdavis: Two thoughts:

First, I lean toward B. I don't think core should have anything it's not using, and I think it's totally fine for www to build atop what it can bring from core. That said, because we're adding more, automating this process then feels a little goofy, since it'd take manual intervention from www anyways.

That's why I end up on B. Though it's a little painful that it feels like we're going more manual. But at this stage of the game, that feels fine.

Second, the structure of `example` in this object is painful. It looks tedious to put together and prone to errors. I'd suggest making it an object, then stringifying it and displaying how you'd like on the front end.

seancdavis (commented):
Added to the inline comments. Overall, I'm loving the additions here! Nice work!

data/property-ops-dictionary--postgres.json
teallarson: @evantahler per @seancdavis ' comment, B sounds like a logical route for me to go unless you have any other thoughts?

evantahler: 👍 for "B" from me! 🐝

teallarson: @seancdavis  doing the stringified JSON tidied things up quite a bit, thanks!

teallarson:
PR for removing the static refs (will add to description as well) https://github.com/grouparoo/grouparoo/pull/1889

evantahler (approved):
Woo!

PR #482, opened 2021-05-28
Add BigQuery reference doc.

pauloouriques:


data/plugins/googleBigQuery.ts
```diff
@@ -1,7 +1,7 @@
 import { Plugin } from "./index";
 
 export const GoogleBigQuery: Plugin = {
-  name: "Google BigQuery",
```
evantahler, line 4: @andyjih this would change the name of the plugin on the other pages as well.  What do you think?

evantahler (commented):
👍 from me, but let's wait for @andyjih on the name change

data/plugins/googleBigQuery.ts
pauloouriques: I thought the same thing, but I looked around and all looks fine to me.

evantahler:
@andyjih I'm merging this in with the name change - we can revert if it's a problem

PR #522, opened 2021-06-12
Cloud Hosted signup flow

teallarson:
[This telemetry PR](https://github.com/grouparoo/telemetry/pull/250) should be merged first to load the `TrialRequests` model.

This PR:
- Adds a page at `/get-started` that is now linked to from the CTAs
- Adds a Trial Request form at `/trial` that feeds to a new `TrialRequest` model in telemetry
- Updates the `/pricing page` (cards and table)

### DIFF
--- a/components/home/getStarted.tsx
+++ b/components/home/getStarted.tsx
@@ -15,33 +15,11 @@ export default function GetStarted({
               <br />
               <br />
               <h2>Get Started with Grouparoo</h2>
-              <p>
-                Easily install the Grouparoo application, join the community, or
-                schedule a demo to learn more.
-              </p>
-              <ButtonGroup>
-                <Button
-                  variant="primary"
-                  href="/docs/getting-started"
-                  className="ml-auto"
-                >
-                  Try it now
-                </Button>
-                <Button
-                  variant="outline-primary"
-                  href="/docs/community"
-                  className="ml-auto"
-                >
-                  Join the Community
-                </Button>
-                <Button
-                  variant="outline-primary"
-                  href="/meet"
-                  className="ml-auto"
-                >
-                  Get a Demo
-                </Button>
-              </ButtonGroup>
+              <p>Start syncing your data with Grouparoo Cloud</p>
+
+              <Button variant="primary" href="/trial" className="ml-auto">
+                Start Free Trial
+              </Button>
 
               {showLearnMore ? (
                 <>
@@ -49,14 +27,10 @@ export default function GetStarted({
                   <br />
                   <br />
                   <p>
-                    Or learn more about the{" "}
-                    <Link href="/about">
-                      <a>company</a>
+                    Or download and try our open source{" "}
+                    <Link href="/docs/getting-started">
+                      <a> Community edition</a>
                     </Link>{" "}
-                    or how to get{" "}
-                    <Link href="/docs/support">
-                      <a>support</a>
-                    </Link>
                     .
                   </p>
                 </>
--- a/components/layouts/floating.tsx
+++ b/components/layouts/floating.tsx
@@ -1,27 +1,12 @@
-import { useState, useEffect, Children, cloneElement } from "react";
+import { Children, cloneElement } from "react";
 import Head from "next/head";
 import GoogleAnalytics from "../googleAnalytics";
 import AdConversionScript from "../adConversionsScript";
-import Navigation from "../navigation";
-import Footer from "../footer";
-import { Alert } from "react-bootstrap";
-import Link from "next/link";
-import { useRouter } from "next/router";
-import GetStarted from "../home/getStarted";
+import { Container } from "react-bootstrap";
+// import { useRouter } from "next/router";
 
-function PageTemplate({ children }) {
-  const router = useRouter();
-  const [releaseNote, setReleaseNote] =
-    useState<{
-      name: string;
-      description: string;
-    }>(null);
-
-  useEffect(() => {
-    if (router.asPath !== "/") setReleaseNote(null);
-  }, [router.asPath]);
-
-  const noGetStarted = ["/meet", "/docs", "/blog", "/jobs", "/get-started"];
+function FloatingLayout({ children }) {
+  // const router = useRouter();
 
   return (
     <>
@@ -128,27 +113,14 @@ function PageTemplate({ children }) {
         <meta name="viewport" content="width=device-width" />
         <link rel="stylesheet" href="/css/prism.css"></link>
       </Head>
-
-      {releaseNote?.description ? (
-        <Alert variant="primary" style={{ textAlign: "center", margin: 0 }}>
-          <small>
-            <strong>What's New</strong> - {releaseNote.description}{" "}
-            <Link href="/whats-new">
-              <a>Learn More</a>
-            </Link>
-          </small>
-        </Alert>
-      ) : null}
-
       <div className="main">
-        <Navigation />
-        <div id="content-container" style={{ minHeight: 700 }}>
-          {Children.map(children, (child) =>
-            cloneElement(child, { setReleaseNote })
-          )}
-        </div>
-        {noGetStarted.includes(router.asPath) ? null : <GetStarted />}
-        <Footer />
+        <Container
+          fluid
+          className="align-items-center justify-content-center d-flex my-5 my-xl-0 pb-lg-3 pb-xl-0 mx-0 trialContainer"
+        >
+          {" "}
+          {children}
+        </Container>
       </div>
 
       <GoogleAnalytics />
@@ -157,4 +129,4 @@ function PageTemplate({ children }) {
   );
 }
 
-export default PageTemplate;
+export default FloatingLayout;
--- a/components/layouts/floating.tsx
+++ b/components/layouts/floating.tsx
@@ -3,11 +3,8 @@ import Head from "next/head";
 import GoogleAnalytics from "../googleAnalytics";
 import AdConversionScript from "../adConversionsScript";
 import { Container } from "react-bootstrap";
-// import { useRouter } from "next/router";
 
 function FloatingLayout({ children }) {
-  // const router = useRouter();
-
   return (
     <>
       <Head>
@@ -16,8 +13,6 @@ function FloatingLayout({ children }) {
         )}
 
         <title>Grouparoo</title>
-        {/* <link rel="stylesheet" href="/css/bootstrap.min.css" />
-        <link rel="stylesheet" href="/css/prism.css" /> */}
         <meta name="application-name" content="Grouparoo" />
         <link
           rel="apple-touch-icon-precomposed"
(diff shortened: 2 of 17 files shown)


pages/get-started.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <div
+        id="headline"
+        className="homePageSection"
+        style={{ textAlign: "center", paddingTop: 0, paddingBottom: 20 }}
+      >
+        <Container>
+          <Row className="pt-5 pb-3">
+            <Col>
+              <h1>Sync data how you want.</h1>
+              <h2>Lorem ipsum</h2>
```
teallarson: @andyjih found one more spot if you want to add this to your commit!

evantahler:
Suggestion: Perhaps `/docs/getting-started` should be renamed now that we have `/get-started`... perhaps `/docs/welcome` or something like that.

evantahler:
@andyjih the pricing page still doesn't have any prices 💵

components/home/getStarted.tsx
```diff
@@ -15,48 +15,22 @@ export default function GetStarted({
...
-                  className="ml-auto"
-                >
-                  Get a Demo
-                </Button>
-              </ButtonGroup>
+              <p>Start syncing your data with Grouparoo Cloud</p>
+
+              <Button variant="primary" href="/trial" className="ml-auto">
+                Start Free Trial
+              </Button>
```
evantahler, line 22: Nice! A much clearer flow

components/solutions/comparisonPage.tsx
```diff
@@ -49,8 +49,8 @@ export default function ComparisonPage(comp: ComparisonInfo) {
 
           <Row>
             <Col>
-              <Button variant="primary" href="/meet" className="ml-auto">
-                Get a Demo →
+              <Button variant="primary" href="/get-started" className="ml-auto">
+                Start Free Trial
```
evantahler, line 53: I imagine the same should be changed on the `/solutions/*` pages too? They still point to `/meet`

pages/index.tsx
```diff
@@ -68,36 +68,19 @@ export default function IndexPage({ pageProps, setReleaseNote }) {
...
-                  height="315"
-                  src="https://www.youtube.com/embed/kQ789gMXJB8?rel=0"
-                  frameBorder="0"
-                  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
-                  allowFullScreen
-                ></iframe>
-              </p>
-            </Col>
-            <Col lg={2} />
-          </Row>
```
evantahler, line 99: 👋 🎥

pages/pricing.tsx
```diff
@@ -277,6 +276,26 @@ export default function PricingPage() {
                     />
                   </td>
                 </tr>
+                <tr>
+                  <td style={{ textAlign: "left" }}>SaaS Hosted</td>
```
evantahler, line 281: @andyjih is using the word "SaaS" OK?  I thought we wanted to use "Hosed by Grouparoo" or something like that?

andyjih:
> @andyjih the pricing page still doesn't have any prices 💵

updated!

pages/pricing.tsx
```diff
@@ -77,67 +75,69 @@ export default function PricingPage() {
...
+                 shadow-lg h-100"
                 border="primary"
                 bg="secondary"
                 text="white"
               >
                 <Card.Body>
-                  <Card.Title>Self-Managed Enterprise Edition</Card.Title>
-                  Empowers non-technical users and teams to use Grouparoo.
+                  <Card.Title>Standard Cloud</Card.Title>
+                  <h5>$150/month</h5>
```
evantahler: Starting at $150/mo?

pages/trial.tsx
```diff
@@ -0,0 +1,372 @@
...
+                    </div>
+                  </Form.Group>
+                  <Form.Group>
+                    <div className="my-3">
+                      <Form.Check
+                        {...register("check", { required: true })}
+                        type="checkbox"
+                        id="privacy-policy-check"
+                        defaultValue="false"
+                        label={privacyPolicyLabel}
```
evantahler: Perhaps the label should be centered and in a smaller font?

Screen Shot 2021-06-15 at 4 08 28 PM

pages/trial.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+    data.subdomain = subdomain;
+    data.source = source;
+    data.medium = medium;
+    data.campaign = campaign;
+
+    // TO DO: CONFIGURE GOOGLE ANALYTICS CONVERSION DATA
+    if (isBrowser() && globalThis?.gtag) {
+      globalThis.gtag("event", "conversion", {
+        send_to: "send_to': 'AW-467110449/dWjCCL2x78ICELGU3t4B", // demoRequest conversion
```
evantahler: This is likely a new adWords conversion event, and should have a new ID (cc @andyjih to make the new event ID)

pages/trial.tsx
```diff
@@ -0,0 +1,372 @@
...
+                        required: true,
+                        pattern: urlPattern,
+                      })}
+                      type="text"
+                      name="companyWebsite"
+                      defaultValue=""
+                      onChange={onSubdomainChange}
+                      disabled={disabled}
+                    />
+                  </Form.Group>
```
evantahler: Perhaps... we could just get the subdomain from the email address?  If I'm `evan@foo.com` it seems like a likely guess that the subdomain I want is `foo.grouparoo.app`?

evantahler: In general, I think the validation in use here is too strong anyway.  The subdomain didn't appear until I added `www` in front of `www.company.com`.   

Screen Shot 2021-06-15 at 4 11 29 PM

pages/trial.tsx
```diff
@@ -0,0 +1,372 @@
...
+                  <div className="mt-3 mb-5">
+                    <Row className="my-3">
+                      <Col>
+                        <p className="h4">Getting Started Video</p>
+                        <a
+                          href="https://www.youtube.com/embed/kQ789gMXJB8?rel=0"
+                          target="_blank"
+                        >
+                          A quick intro into the Grouparoo platform
+                        </a>
```
evantahler: I'd suggest taking the user to `/docs/getting-started` instead - it's got the video and more info

package.json
```diff
@@ -33,6 +33,7 @@
     "babel-plugin-transform-define": "2.0.0",
     "bootstrap": "4.6.0",
     "cheerio": "1.0.0-rc.9",
+    "extract-domain": "^2.3.3",
```
evantahler: Might not be needed if you move route of getting the domain from the email address

pages/_app.tsx
```diff
@@ -45,6 +45,14 @@ export default function GrouparooWWW(props) {
...
   }
 
+  if (router.pathname === "/trial") {
+    return (
+      <>
+        <Component {...props} />{" "}
+      </>
+    );
+  }
+
```
evantahler: I would suggest making a new layout.  We have `main`, and you just created `floating` or something like that.  You can extract the container, error display, and a few other parts from the `trial` page into a new `components/layouts/floating` and use that.  Then, this section becomes something like:

```tsx
const LayoutComponent = router.pathname === "/trial" ? MainLayout : FloatingLayout
return <LayoutComponent><Component {...props} /></LayoutComponent>
```

evantahler (changes requested):
I really like these new pages!  They are clear, have a linear flow that makes sense, and are easy to understand!

I think there are 2 main areas to clean up:
* I think there is a new Layout created here. which can be extracted from the page
* I think the logic to get the customer's subdomain is a bit to smart, and I'd suggest extracting it from the customer's email address

pages/trial.tsx
andyjih: @evantahler That is the new ID. I think the comments are just out of date.

teallarson: Oops!  Yes I pasted in the new id and did not update the comment.

pages/trial.tsx
teallarson: Yeah it definitely needs a change!  Andy pointed it out too.  Thoughts on using the email instead of the url to create the domain name @andyjih ?  

I imagine our target user isn't out signing up for things with an '@gmail' or '@outlook' email, but that would be my only concern there... though it's not a big one (people can change it, and the serverside validation would catch duplicates)

andyjih: We have had some people reach out for demo requests with @gmail email addresses, and I generally want to filter those out if possible. So I guess the rough options as I see it are:
1. Use the email domain, but add validation to exclude common personal email domains (gmail, yahoo, aol, outlook, etc.)
2. Keep using company website, but loosen the validation to allow URLs that don't have `www` or `http`.

I lean slightly towards 2, but definitely open to either option!

teallarson: I'm leaning towards 2 given what you shared about demo requests

pages/pricing.tsx
andyjih: @evantahler I think SaaS feels right to me here. Just using the word "Hosted" suggests they're only paying for hosting, which isn't entirely true.

components/layouts/floating.tsx
```diff
@@ -0,0 +1,132 @@
+import { Children, cloneElement } from "react";
+import Head from "next/head";
+import GoogleAnalytics from "../googleAnalytics";
+import AdConversionScript from "../adConversionsScript";
+import { Container } from "react-bootstrap";
+// import { useRouter } from "next/router";
```
evantahler: Some comments to clean up

pages/trial.tsx
```diff
@@ -0,0 +1,391 @@
+import { Container, Button, Row, Col, Form, Image } from "react-bootstrap";
+import Head from "next/head";
+import { useForm } from "react-hook-form";
+import Link from "next/link";
+import { useState } from "react";
+import { useApi } from "./../hooks/useApi";
+import { ErrorHandler } from "../utils/errorHandler";
+// import extractDomain from "extract-domain";
```
evantahler: more comments :D

pages/trial.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+    data.subdomain = subdomain;
+    data.source = source;
+    data.medium = medium;
+    data.campaign = campaign;
+    data.email = data.email.toLowerCase();
+    data.companyWebsite = data.companyWebsite.toLowerCase();
+    data.subdomain = data.subdomain.toLowerCase();
+
+    // TO DO: CONFIGURE GOOGLE ANALYTICS CONVERSION DATA
```
evantahler: I think this TODO is done?

pages/trial.tsx
```diff
@@ -0,0 +1,391 @@
...
+    <>
+      <Head>
+        <title>{title}</title>
+        <link rel="canonical" href="https://www.grouparoo.com/trial" />
+      </Head>
+
+      {/* <Container
+        fluid
+        className="align-items-center justify-content-center d-flex my-5 my-xl-0 pb-lg-3 pb-xl-0 mx-0 trialContainer"
+      > */}
```
evantahler: Unused

evantahler (approved):
Great work! 

Just some comments to remove/update and this is ready to go!

The telemetry server (https://github.com/grouparoo/telemetry/pull/250) has been deployed so the form works!

PR #524, opened 2021-06-12
Revert "`head_ref` does not work in forks"

evantahler:
Reverts grouparoo/www.grouparoo.com#511

This IS needed to the auto-update action knows what branch to push things to.  

This will bring back problems for people who try to contribute on external forks, but we'll have to find another solution


evantahler:
cc @andyjih

PR #532, opened 2021-06-17
error on `/meet` form if no email

teallarson:
Previously, React Hook Form would block form submission with a missing email, but no error would display (because the errors were wired to come from the server).  I've added React Hook Form error handling for the email field similar to on `/trial`.

Now, users get an error: 
Screen Shot 2021-06-17 at 3 19 36 PM

### DIFF
--- a/pages/meet.tsx
+++ b/pages/meet.tsx
@@ -13,13 +13,18 @@ interface FormError {
   generic: string;
 }
 export default function Meet() {
-  const { handleSubmit, register } = useForm();
+  const {
+    handleSubmit,
+    register,
+    formState: { errors },
+  } = useForm();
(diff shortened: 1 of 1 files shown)


evantahler (approved):
It's a little odd that some of the validations appear in the new red text below the field (try ""), but others appear in the browser pop up (try "fff").  

But... this is better than it was! 

<table>
<tr>
<td>
Screen Shot 2021-06-17 at 6 13 27 PM
</td>
<td>
Screen Shot 2021-06-17 at 6 13 34 PM
</td>
</tr>
</table>

teallarson:
The pop up one is browser validation. I'll find how to turn that off. Thanks, but thanks, Chrome.

PR #533, opened 2021-06-22
UI Config Docs

seancdavis:
⚠️ **This should not be merged until we're ready to release UI Config.** ⚠️ 

This introduces several changes to our docs section in support of UI Config. The detailed changes can be found [here](https://docs.google.com/document/d/1pkda0liBoEV8wo5O4Qu-5OfyRBxSluhYKb82Phxl09Y/edit#). But here's a brief summary:

- Enterprise guides moved to /docs/running to be more accurate (they can only be used when actually _running_ Grouparoo).
- Code config guides have been hidden from the nav.
- There are now primary configuration guides under /docs/config/[thing] for each of the six _things_ in UI Config.
- Updates to various landing pages to support new language that UI Config is the recommended way to configure your Grouparoo application.
- Add/remove redirects, as appropriate.
- Updated nav.
- Placed screenshot images (including old ones in Enterprise guides) in browser frame.
- Updated references to existing docs to account for new routes.

PR #561, opened 2021-07-01
v0.5 Migration Guide

seancdavis:
⚠️ (This expects #533 to be merged first.) ⚠️ 

After thinking about this, it seems like there aren't any breaking changes. This is just a new way to generate those files.

Could have added more details, but this seemed like enough since it points back to the new Config UI docs, which have more detail.


evantahler:
Update guide updated to include the removal of the Manual app (breaking change)

PR #562, opened 2021-07-02
[SEO] meta tags/h1 updates

teallarson:
- Edits per feedback from Victorious SEO.
- Added a meta title and description field to the category page .mdx files so we can customize them more smoothly in the future.


components/home/integrationsHead.tsx
```diff
@@ -12,10 +12,10 @@ export default function IntegrationsHead({
   return (
     <>
       <Head>
-        <title>{name} Integrations - Best Data App Integration Tools</title>
+        <title>{name} Data Integration ETL Tools - Import &#38; Upload</title>
```
andyjih: I actually have a question out with the SEO agency about this one. They made general recommendations to ~all integrations pages which I already merged, but this Salesforce one was an outlier. Mind if we revert this change for now until they get back to me? Sorry about that!

pages/integrations/[type]/[plugin].tsx
```diff
@@ -196,7 +196,9 @@ function DestionationHeaderArrow({ data }) {
       <Row>
         <Col lg={5} md={12}>
           <br />
-          <h1>{data.name} integration for all your tools & customer data</h1>
+          <h1>
```
Comment (andyjih) [inline pages/integrations/[type]/[plugin].tsx]: Let's hold off on this change as well. Sorry!

andyjih (changes requested):
Category Page changes look great. Let's pause on the Integrations pages change for a minute.

evantahler (approved):
After @andyjih's changes, fine with me!

components/home/integrationsHead.tsx
```diff
@@ -15,7 +15,7 @@ export default function IntegrationsHead({
         <title>{name} Integrations - Best Data App Integration Tools</title>
         <meta
           name="description"
-          content={`The Grouparoo ${name} integration allows you to sync all of your customer data that lives in various sources into ${name}. ✓ Start a free trial now!`}
+          content={`Grouparoo\'s ${name} integration allows you to sync all of your customer data that lives in various sources into ${name}. ✓ Start a free trial now!`}
```
andyjih: @tealjulia The apostrophe shows up as below: 
Screen Shot 2021-07-06 at 4 10 40 PM

Evan and I spent a bit of time looking into it and decided the copy change of "The Grouparoo ${name}" gets the job done without the escaping challenges. Thoughts on reverting this change?

andyjih (changes requested):
Just one small change, but open to discussion.

components/home/integrationsHead.tsx
teallarson: Yeah no problem

andyjih (approved):

PR #572, opened 2021-07-12
Noindexed tag and author pages, whats new pages

teallarson:
Noindexing these pages means they will not be crawled and will not show up in search results.

The reasoning behind this is that the thin content on them drags the other pages down.  Noindexing them will boost the other pages that we want people to be directed to.

Pages affected: 
- feeds by author and tag
- individual what's new posts

Pages unaffected:
- what's new feed
- blog feed
- blog posts

### DIFF
--- a/components/seo.tsx
+++ b/components/seo.tsx
@@ -37,6 +37,7 @@ const SEO = ({
   twitterAuthor,
   twitterCard,
   path,
+  noIndex = false,
 }: {
   canonical?: boolean;
   description?: string;
@@ -45,6 +46,7 @@ const SEO = ({
   twitterAuthor?: string;
   twitterCard?: string;
   path: string;
+  noIndex?: boolean;
 }) => {
   if (!path || !title) {
     console.error(`path and title are required for SEO component.`);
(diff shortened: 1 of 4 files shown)


evantahler (commented):
A few questions that might be relevant:
* If these pages are excluded from being crawled, does that mean that it will be hard to discover that we have RSS feeds?  I'm thinking of apps like Feedly that might also have crawlers for RSS content. 
* If these pages are both in the sitemap and then say `noindex` on them, who wins?  The sitemap or the page?

teallarson:
> A few questions that might be relevant:
> 
> * If these pages are excluded from being crawled, does that mean that it will be hard to discover that we have RSS feeds?  I'm thinking of apps like Feedly that might also have crawlers for RSS content.
> * If these pages are both in the sitemap and then say `noindex` on them, who wins?  The sitemap or the page?

Noindexed pages should not be in the sitemap it looks like (https://sitebulb.com/hints/xml-sitemaps/noindex-url-in-xml-sitemaps/)

evantahler (approved):

teallarson:
@evantahler good catch! I changed it to only affect the author and category feeds, not the main one.  It does look like it would interfere with the RSS feeds though.  

There is an option to only exclude the Google crawlerbot by changing the meta tag wording, which I'm guessing would not exclude it from tools like Feedly.

andyjih (approved):

PR #574, opened 2021-07-14
SEO - add gtags for goal events

teallarson:
To configure the goals laid out for us by Victorious, we need to track specific user events via gtags.  This PR adds code to send event gtags upon:
- submission of a Trial Request
- submission of a Demo Request

Edit: conversions for joining our community are being configured by the generic gtag and are not triggered by an event tag.


pages/chat.tsx
```diff
@@ -31,7 +47,7 @@ export default function Chat() {
       </div>
 
       <div id="signup" style={{ textAlign: "center" }}>
-        <Container style={{ maxWidth: 880 }}>
+        <Container style={{ maxWidth: 880 }} onClick={() => onJoinClick()}>
```
evantahler: This is a really fascinating problem - how do we instrument an iframe we don't have control over!   

Good call keeping track of if the event has been sent already (they will potentially click a few times).  Personally, I'm a keyboard-first browser, and I'll tab into the form... Will some of the other events work here?  `onKeyDown` for instance?

Also, are there any APIs/etc available from Comsor for them to fire off the tracking event for us?

pages/trial.tsx
```diff
@@ -114,6 +114,9 @@ export default function Trial({ props }) {
         send_to: "send_to': 'AW-467110449/dWjCCL2x78ICELGU3t4B", // trialRequest conversion
         event_callback: () => {},
       });
+      globalThis.gtag("event", "trial-request", {
+        event_category: "conversion",
+      });
```
evantahler: The other gTag events have `event_callback`... Should that be added to these new events, or perhaps removed from the others?

pages/trial.tsx
```diff
@@ -114,6 +114,9 @@ export default function Trial({ props }) {
         send_to: "send_to': 'AW-467110449/dWjCCL2x78ICELGU3t4B", // trialRequest conversion
```
evantahler: This addwords account should probably be moved to an environment variable

teallarson: That sounds right.  I believe there is at least one elsewhere as well from setting up the paid search tracking.

pages/trial.tsx
teallarson: The callback functions are empty so I don't think they're actually doing anything?

Here's the reference in the gtag.js docs https://developers.google.com/analytics/devguides/collection/gtagjs/sending-data

tealjulia: Suppose I can try removing them and see if we still get data for our ads? 😅

pages/chat.tsx
teallarson: Commsor does have an API... but it doesn't do what we want.

I'm going with the idea from standup to just track visits to the page versus actual interaction with the page and updating this goal for now.

teallarson:
After the above comments:
- Removed tracking for the iform, set the goal based on page view.
- Moved the Google Ads tracking to an env variable

pages/trial.tsx
```diff
@@ -114,6 +114,10 @@ export default function Trial({ props }) {
...
```
evantahler: Looks like one more tracking variable to move to ENV

evantahler (changes requested):
* It looks like this PR lost the `/chat` conversion?  Is the goal instrumented just by a google analytics pageview?
* Looks like there is one more google account to move to ENV.

teallarson:
> * It looks like this PR lost the `/chat` conversion?  Is the goal instrumented just by a google analytics pageview?

the `/chat` conversion is now based on a page view, so it no longer needs its own tag.  Got the other tag added to env and switched them to production-only so we aren't tracking ourselves filling out forms on previews.

evantahler (approved):

PR #576, opened 2021-07-16
only send gtag if conversion successful

teallarson:
Per discussion in planning this morning, this PR moves the event tag trigger to only happen when there is a successful response from the server for a demo or trial request.


andyjih (approved):
Looks good to me!

PR #580, opened 2021-07-19
SEO - update to page title lengths

teallarson:
Page titles are now all shorter than 60 chars (except What's New posts, which are `noindex`ed so don't affect SEO)
Most are in the 55-60 char range.

There is also now a test for page title length that keeps us under 60 characters.  No minimum yet -- many page titles are still fairly short, and there's no clear minimum.

The main place this will impact us going forward is watching blog post lengths.  60 chars (including ": Grouparoo Blog") goes fast.


scripts/generate_plugin_json.ts
```diff
@@ -28,7 +28,7 @@ if (!fs.existsSync(manifestFileDir)) {
 }
 
 const localPluginPaths = docsNavData
-  .find((plugin) => plugin.title === "Plugins")
+  .find((plugin) => plugin.title.includes("Plugins"))
```
teallarson, line 31: Because we lengthened "Plugins" titles to "Installing Grouparoo Plugins" and "Plugins Documentation", we have to change how this is generated.

__tests__/integration/html.ts
```diff
@@ -79,12 +79,19 @@ describe("sitemap integration", () => {
...
         const title = $("title").text();
         expect(title).toBeTruthy();
       });
+      test("SEO: title length less than 60 chars", async () => {
+        //what's new posts are noindexed so it's ok if they're longer
+        if (!testUrl.includes("whats-new")) {
+          const title = $("title").text();
+          expect(title.length).toBeLessThanOrEqual(60);
+        }
+      });
```
evantahler, line 92: 🎉

data/docs-nav.ts
```diff
@@ -23,9 +23,18 @@ const DocsNav: NavItem[] = [
...
+      },
+      {
+        title: "Installing Grouparoo with Docker",
+        path: "/docs/installation/docker",
+      },
+      {
+        title: "Installing Grouparoo Plugins",
+        path: "/docs/installation/plugins",
+      },
     ],
```
evantahler: I understand the desire to have the titles here match the page titles... but now the nav wraps which isn't great for legibility.  Can these be shorter and not match the page titles?

Screen Shot 2021-07-20 at 2 14 40 PM

evantahler: Similar comments below

evantahler (commented):
I'm glad that the title change doesn't effect the slugs, so we don't need to add any redirects!

data/docs-nav.ts
tealjulia: Good point.  I added an optional "linkText" parameter for places we want the link text to be different on those.

teallarson: Good point.  I added an optional `linkText` property now.

data/docs-nav.ts
```diff
@@ -221,19 +246,24 @@ const DocsNav: NavItem[] = [
...
-      { title: "Secrets", path: "/docs/support/secrets" },
-      { title: "Security", path: "/docs/support/security" },
-      { title: "Telemetry", path: "/docs/support/telemetry" },
+      {
+        title: "Grouparoo Privacy Policy & Legal",
+        linkText: "Privacy Policy & Legal",
+        path: "/docs/support/legal",
+      },
+      { title: "Support Docs: Monitoring", path: "/docs/support/monitoring" },
+      { title: "Suport Docs: Network", path: "/docs/support/network" },
```
andyjih:
```suggestion
      { title: "Support Docs: Network", path: "/docs/support/network" },
```

pages/blog/any-wreaks-havoc-on-overloads.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: "How TypeScript's any creates confusing function overload bugs"
+title: "How TypeScript's any creates confusing bugs"
```
andyjih:
```suggestion
title: "How TypeScript's `any` creates confusing bugs"
```

pages/blog/getting-previous-path-nextjs.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Getting the Previous Path in Next.js Application"
+title: "Get User's Previous Path with NextJS Router"
```
andyjih:
```suggestion
title: "Get user's Previous Path with NextJS Router"
```

Not sure if User should be capitalized or not?

andyjih (changes requested):
Overall looks great! Just a few modifications.

PR #587, opened 2021-07-21
SEO/SEM - Sitemap exclude noindex

teallarson:
Now that our `whats-new` posts are not indexed, we should not include them in our sitemap.  Google doesn't like it.

Given that there are so few categories of pages that we want to `noindex`, I've gone with just manually excluding them.  There may be a more direct way to do this in the way we don't add our category/author feeds to our sitemap, though I haven't been able to get that to work.  It may make sense in the future to use a different npm package to build our site index that does crawl meta tags. 

Three main changes:
- Removed whats-new pages from sitemap manually
- Excluded them explicitly from future sitemap builds
- Adjusted testing to factor that in


evantahler (approved):
I think the approach of filtering our `/whats-new` pages from the sitemap makes sense!  Can you please add a test to `__tests__/integration/sitemap.ts` to ensure that there are no what's new pages?

andyjih (approved):
lgtm!

teallarson:
> I think the approach of filtering our `/whats-new` pages from the sitemap makes sense! Can you please add a test to `__tests__/integration/sitemap.ts` to ensure that there are no what's new pages?

✅

PR #588, opened 2021-07-22
Add narrative section to integrations

teallarson:
We're starting with Salesforce, Zendesk, and Marketo as those are our SEO focus integrations.

I've added a new section for a narrative so we can fit search-related keywords onto these pages to drive traffic.  

Also, thought, should we have a CTA higher on the page?  I think we want users to be able to act on information faster.   Below where we link to the Community/Enterprise docs might be a good place to link to a trial.


data/plugins/marketo.ts
```diff
@@ -58,4 +58,8 @@ export const Marketo: Plugin = {
   ],
   otherPluginsHeading:
     "Import your customer data from these tools into Marketo",
+  narrative: {
+    p1: "Data is the center of your business, let us help you set yourself up for success.  Connect your data warehouse with Marketo using Grouparoo.  Grouparoo enables you to read data from multiple resources and sync it straight to Marketo for your marketing team to access. No overbuilt systems or spending time and engineering resources to build your own data management pipeline.  Our application features over 20 integration plugins including our Google BigQuery integration and our Postgres database integration, so you can connect all your external applications to gather a singular view of your data -- even if your tools change over time.",
```
evantahler, line 62: * read data from multiple sources

evantahler (approved):
👍 from me, I'll let @andyjih weigh in on the content

andyjih (approved):
Nice work!

PR #589, opened 2021-07-23
Cannibalizing content fix

teallarson:
For now, I've pulled the content.  

Originally, we'd thought the content should move to `/integrations`, but it didn't seem to add much value there as it was pretty implementation-specific.  It would be more helpful in either the plugin docs or as standalone tutorials.

I did save the content, and images, on Google Drive so when we are ready to write those we have access to it!


evantahler (commented):
I'm OK with removing the posts we don't want from the project, if you don't think we'll want that content in the future (cannibalizing).  

If you did want to keep the content around for a little bit but not have it display, you could add a new property to the frontmatter (eg: `display: false`) and have it not be loaded up by the `Blog.getBlogEntries` method and related methods.

teallarson:
> I'm OK with removing the posts we don't want from the project, if you don't think we'll want that content in the future (cannibalizing).
> 
> If you did want to keep the content around for a little bit but not have it display, you could add a new property to the frontmatter (eg: `display: false`) and have it not be loaded up by the `Blog.getBlogEntries` method and related methods.

Interesting.  I think we _may_ want it later (for docs or tutorials), but not for sure.  I do have the content backed up on Google Drive and can add it to a shared folder as well. 

But I don't think it's worth re-adding and adding an additional property to the frontmatter.

andyjih:
Overall this makes sense! Should we put in redirects for the pages that don't exist anymore?

teallarson:
> Overall this makes sense! Should we put in redirects for the pages that don't exist anymore? 

Hmm possibly if someone finds it through a search or back link. That sounds smart. There are no internal links anymore to these posts. I guess we'd use permanent redirects (301?) since the pages are gone for good?

andyjih:
301 sounds right! Maybe we redirect them to their respective `/integrations` pages?

teallarson:
Screen Shot 2021-07-28 at 7 46 23 AMAlright!  All 301s now @andyjih

evantahler (approved):

PR #593, opened 2021-07-27
add page_path to ga tag

teallarson:
Based on [this article](https://mariestarck.com/add-google-analytics-to-your-next-js-application-in-5-easy-steps/) sent to us from Victorious, I think this is the missing piece.  This was their exact feedback on how our ga tag was not tracking properly:

Screen Shot 2021-07-27 at 9 32 55 AM

### DIFF
--- a/components/googleAnalytics.tsx
+++ b/components/googleAnalytics.tsx
@@ -11,7 +11,7 @@ function GoogleAnalytics() {
 window.dataLayer = window.dataLayer || [];
 function gtag(){dataLayer.push(arguments);}
 gtag('js', new Date());
-gtag('config', '${GoogleAnalyticsTrackingID}');
+gtag('config', '${GoogleAnalyticsTrackingID}', {page_path: window.location.pathname});
 `;
 
     const script = document.createElement("script");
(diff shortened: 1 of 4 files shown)


evantahler (changes requested):
I think this change night not be enough - The article talks about needing to use the router events (step 3+4) to re-fire an event when the SPA (single-page-app) page change.  I think 
 this change likely only effects the initial load of the page.

We already hook into the router events here https://github.com/grouparoo/www.grouparoo.com/blob/main/pages/_app.tsx#L12-L20

teallarson:
@evantahler this seems like what we want.  I also found [this tool](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) that will be helpful for debugging GA in general (I'll check that the pageviews and events are being triggered once this is merged in... I _think_ our events (form submissions) are ok the way they were).

evantahler (approved):
Nice!

teallarson:
Well.  That wasn't it.  But very helpful Chrome extension!
 
Screen Shot 2021-07-28 at 12 06 08 PM

PR #594, opened 2021-07-29
move tracking to _document level

teallarson:
Move GA tracking scripts to the _document level... trying to remove as many layers of abstraction as possible until we find out how to make it actually fire correctly.


pages/_document.tsx
```diff
@@ -4,7 +4,37 @@ class MyDocument extends Document {
   render() {
     return (
       <Html lang="en">
-        <Head />
+        <Head>
+          <script
+            async
+            src={`https://www.googletagmanager.com/gtag/js?id=${process.env.GOOGLE_ANALYTICS_TRACKING_ID}`}
```
evantahler, line 10: Normally you wouldn't want a remote script tag in your `head` because loading it could block the rest of the page showing up... BUT the `async` tag means it doesn't!

evantahler (approved):

PR #595, opened 2021-07-29
Integrations cta

andyjih:
Before:

After:

### DIFF
--- a/components/home/integrationCard.tsx
+++ b/components/home/integrationCard.tsx
@@ -24,6 +24,7 @@ export default function IntegrationCard({
   const card = (
     <div
       style={{ borderRadius: "5px", boxShadow: "5px 5px 5px 5px lightgrey" }}
+      className="h-100"
     >
       <Card
         bg="dark"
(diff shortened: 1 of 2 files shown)


components/home/integrationCard.tsx
```diff
@@ -24,7 +24,6 @@ export default function IntegrationCard({
   const card = (
     <div
       style={{ borderRadius: "5px", boxShadow: "5px 5px 5px 5px lightgrey" }}
-      className="h-100"
```
andyjih: @tealjulia I noticed that the integration card height was getting kinda strange. I removed this classname so that the card content is centered vertically, but I wanted to see if I broke something else. It looks like you added this, so please let me know if this is a bad move on my part or if there's a different way to address it.

teallarson: That does make the integrations page cards non-uniform in size now (https://www-grouparoo-com-git-integrations-cta-grouparoo.vercel.app/integrations).  

I'm sure there's a way to use grid to make it all right.  I bet they're using flexbox right now.

teallarson: Other option:  The h-100 class makes it 100% the height of the parent container, so if you explicitly set the parent container on the individual integration page (the one you screenshotted above), I think it could do the same thing, but not sure how that would look across devices.

PR #606, opened 2021-08-02
add docs for how to access REST api

teallarson:
Generating and obtaining an API key (in UI or code config) and how to find docs for what the API can do.


teallarson:
Still missing because I'm unclear.  If a user generates an API key in code config... they have to either have an existing API key to call `GET` on `/apiKey/{id}` OR open the UI to view it there... correct?

pages/docs/support.mdx
```diff
@@ -42,6 +42,7 @@ We also have an array of support topics documented to help you learn more about
 - [Monitoring](/docs/support/monitoring)
 - [Network](/docs/support/network)
 - [Permissions](/docs/support/permissions)
+- [REST API](/docs/support/rest-api)
```
evantahler, line 45: Looks like this also needs to be added to the navigation

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+## What can you do with our API?
+
+Grouparoo's REST API contains endpoints for over 25 topics -- everything from installing a Grouparoo Plugin to getting a list of all your Schedules. To see a full list of API endpoints, navigate to `/swagger` in the UI of your Grouparoo instance or run:
+
+```bash
+curl -X 'GET' \
+  '{{yourinstanceurl}}/api/v1/swagger' \
+  -H 'accept: application/json'
+```
```
evantahler: I think it's a little odd to use the Swagger endpoint to learn about the API.  I like telling folks to visit `/swagger` in the UI.  We can also suggest they use a tool like Postman (it can import swagger)?

Screen Shot 2021-08-03 at 5 38 20 PM

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,110 @@
...
+      id: "{{yourAPIKeyId}}",
+      name: "{{yourAPIKeyId}}",
+      options: {
+        permissionAllRead: false, // Does this API Key have read permission to all topics?
+        permissionAllWrite: false, // Does this API Key have write permission to all topics?
+      },
+    },
+  ];
+};
+```
```
evantahler: Oh! This template should have been updated as part of https://github.com/grouparoo/grouparoo/pull/2117!  I made a story for it, that's probably blocking this now - https://www.pivotaltracker.com/story/show/179111715

My fault!

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+From here, you can update the name and permissions.
+
+### Accessing the API
```
evantahler: > ### Accessing & Testing the API

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,110 @@
...
+
```
evantahler, line 64: We should add something like:

> As Grouparoo can be self-hosted, every deployment of Grouparoo provides its own API documentation.

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,110 @@
...
+<ImageInBrowserFrame
+  centered
+  source="Generate an API Key"
+  alt="Generate an API Key"
+  src="/images/docs/support/rest-api/swaggerdocs.png"
+  width={820}
+  height={530}
+/>
+
+To see examples of how to use your API key in the Swagger docs, click "Authorize" and enter your API Key (if pasting in your API Key, be careful about extra whitespace).
```
evantahler: > The Grouparoo API can be authorized in 2 distinct ways: as a web user (with a cookie and CSRF token), or via an API Key.  By default, when visiting the `/swagger` page, you will be authenticated using your existing cookie, but you can switch to using an API key.  To try your new APIKey, click "Authorize" and enter your API Key (if pasting in your API Key, be careful about extra whitespace).

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,110 @@
...
+  height={521}
+/>
+
+For example, if you want to see a list of all your instance's apps, you can run:
+
+```
+curl -X 'GET' \
+  '{{yourInstanceURL}}/api/v1/apps?limit=100&offset=0&apiKey={{yourAPIKey}}' \
+  -H 'accept: application/json'
+```
```
evantahler: This part is actually the most exciting, as it is the demo of what the end user will be doing.   

You may want to experiment with a more full bash script to make this more readable:

```bash
# Set Variables
GROUPAROO_HOST="http://localhost:3000"
GROUPAROO_API_KEY="ba05b18d9dae4acbb08799e7ef238f0d"

# Make Requests
curl -X GET "$GROUPAROO_HOST/api/v1/apps?apiKey=$GROUPAROO_API_KEY"
```

evantahler: I would love to see a workflow here - perhaps 
1. create a team
2. add someone to that team

evantahler (changes requested):
There's certainly a missing part in the generator that I forgot (sorry!) - see the comments below and https://www.pivotaltracker.com/story/show/179111715

pages/docs/support/rest-api.mdx
teallarson: With that flow, it feels iffy to me that someone's password would be in their cli history in that way.  Is that being overly paranoid?  I suppose since it's HTTPS the request url itself is encrypted.

evantahler: that's a very good point.  Is there another workflow that might be good?  Maybe adding a source + property to an existing app?

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,189 @@
...
+  source="Generate an API Key"
+  alt="Generate an API Key"
+  src="/images/docs/support/rest-api/auths.png"
+  width={820}
+  height={521}
+/>
+
+For example, if you want to see a list of all your instance's apps, you can run:
+
+```
```
evantahler: try ```bash to get some colors!

evantahler:
```suggestion
```bash
```

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,189 @@
...
+# Set More Variables
+APP_ID="demo_db"
+SOURCE_NAME="demo_source"
+TABLE_NAME="users"
+SOURCE_TYPE="postgres-table-import"
+
+# Make Request
+curl -X POST "$GROUPAROO_HOST/api/v1/source?appId=$APP_ID&name=$SOURCE_NAME&options=%7B%22table%22%3A%20%22$TABLE_NAME%22%7D&type=$SOURCE_TYPE&state=ready&apiKey=$GROUPAROO_API_KEY"
```
evantahler: This request is putting all the variables in the URL and HTML-encoding them... which makes it kind of hard to read.  how about this style:

```bash
# Set new variables for the request
APP_ID="demo_db"
SOURCE_NAME="demo_source"
TABLE_NAME="users"
SOURCE_TYPE="postgres-table-import"

# Make Request
curl \
  -X POST \
  -d "apiKey=$GROUPAROO_API_KEY" \
  -d "appId=$APP_ID" \
  -d "name=$SOURCE_NAME" \
  -d "type=$SOURCE_TYPE" \
  -d "state=ready" \
  -d "options={\"table\": \"$TABLE_NAME\"}" \
  -d "mapping={\"id\": \"user_id\"}" \
"$GROUPAROO_HOST/api/v1/source"
```

* note the trailing `\` which lets you continue the command on a second line in bash (meaning you can paste this into the CLI) 
* Your probably also want to quote your variables
* The example is missing a mapping or a bootstrap maybe?  Or, you could make the source but not put it into the ready state... then make a property... then make it ready?

evantahler (commented):
Getting there!  I think the formatting of the curl command could be easier to read, and I think the example is missing a few steps to get it to work properly

pages/docs/support/rest-api.mdx
teallarson: That's much nicer!  Can you add a property to a source that isn't `"ready"`?  I'll add a mapping!

teallarson:
- Added mapping to source created
- Reformatted curl request to be more readable

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,215 @@
...
+  src="/images/docs/support/rest-api/apikeyconfig.png"
+  width={820}
+  height={502}
+/>
+
+To learn more about what each of the topics can do in the API, you can access the Swagger Docs for the Grouparoo API by navigating to `/swagger` in the UI of your Grouparoo Instance or going to Platform > API Documentation.
+
+#### Code Config Method
+
+You can also generate an API key object with the [`grouparoo generate` command]. To do this, run
```
evantahler: perhaps extra `[` here?

```suggestion
You can also generate an API key object with the `grouparoo generate` command. To do this, run
```

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+<ImageInBrowserFrame
+  centered
+  source="Generate an API Key"
+  alt="Generate an API Key"
+  src="/images/docs/support/rest-api/auths.png"
+  width={820}
+  height={521}
+/>
+
```
evantahler, line 91: I think an bridge sentance is needed here - we are jumping from the swagger UI to the command line.  How about:

```suggestion

You can use the Swagger UI to try out all the APIs available to your Grouparoo instance.  The Grouparoo API is  available to use via any HTTP API Client.  In the following examples, we'll use the ubiquitous [cURL](https://curl.se/) library. 
```

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,215 @@
...
+    },
+    "filters": [],
+    "isArray": false,
+    "createdAt": 1629494788513,
+    "updatedAt": 1629494788588
+  },
+  ...
+}
+```
+
```
evantahler, line 201:
```suggestion
Note how in the above examples, the `apiKey` can be provided either as a GET parameter in the URL, or as form data.  Either way works! 
```

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,215 @@
...
+  -d "key=$PROPERTY_KEY" \
+  -d "type=$PROPERTY_TYPE" \
+  -d "state=ready" \
+  -d "options={\"column\": \"$COLUMN\", \"aggregationMethod": \"$AGGREGATION_METHOD\"}" \
+
+"$GROUPAROO_HOST/api/v1/source"
+
+# Make Request
+
+curl -X POST "http://localhost:3000/api/v1"
```
evantahler: What's this do?  The POST to `"$GROUPAROO_HOST/api/v1/source"` already happened above?

pages/docs/support/rest-api.mdx
```diff
@@ -0,0 +1,215 @@
...
+  -X POST \
+  -d "apiKey=$GROUPAROO_API_KEY" \
+  -d "id=$PROPERTY_ID" \
+  -d "source=$SOURCE_ID" \
+  -d "key=$PROPERTY_KEY" \
+  -d "type=$PROPERTY_TYPE" \
+  -d "state=ready" \
+  -d "options={\"column\": \"$COLUMN\", \"aggregationMethod": \"$AGGREGATION_METHOD\"}" \
+
+"$GROUPAROO_HOST/api/v1/source"
```
evantahler: Should this be a POST to `"$GROUPAROO_HOST/api/v1/property"` ?

evantahler (commented):
Looking a lot better!  I think there are still a few small things to cleanup...

pages/docs/support/rest-api.mdx
teallarson: ah ha, missing link. added.

pages/docs/support/rest-api.mdx
teallarson: copy/paste fail

pages/docs/support/rest-api.mdx
teallarson: copy/paste fail part 2

evantahler (approved):
🥳

PR #609, opened 2021-08-05
Add reference docs for Mongo plugin.

pauloouriques:
- Parameterize components to support the different terminology.


components/docs/plugins/columnarSource.tsx
```diff
@@ -40,11 +45,29 @@ const PluginDocsColumnarSource = ({
...
+                {" "}
+                and for column is <code>{pluginData.columnAlternative}</code>
+              </span>
+            )}
+            . However, we'll keep the <em>Columnar</em> and{" "}
+            <em>Table Source</em> terms since they're part of Grouparoo
+            terminology.
+          </p>
+        </Alert>
+      )}
```
evantahler, line 63: Nice!

components/docs/plugins/querySource.tsx
```diff
@@ -13,7 +13,81 @@ const PluginDocsQuerySource = ({
...
   isTablesUpperCase?: boolean;
 }) => {
-  const slug = plugin.toLowerCase();
+  const pluginData = PluginData.find(({ name }) => name === plugin);
+  const table = pluginData.tableAlternative || "table";
+  const column = pluginData.columnAlternative || "column";
+  const queryLanguage =
+    plugin === "Mongo" ? "MongoDB Query Language (MQL)" : "SQL";
+
+  const querySchedule =
```
evantahler, line 14: Perhaps these code samples should live on the Plugin model?  All the SQL ones can inherit a common string, and Mongo can provide its own.

package-lock.json
```diff
@@ -1,16988 +1,8 @@
 {
```
evantahler: Looks like you need to update your version of NPM  - `npm install -g npm`

evantahler: or... don't check this in :D

evantahler (approved):
This is looking good! I like the parameterization you've added.  Some suggestions below to go even further, but nothing blocking.

teallarson:
Looks good!  

I wonder now that we have Config UI if we want to mention that and link to it somewhere in plugins docs?  Maybe near the top?  It's possible if someone lands on one of these plugin pages from a search that they wouldn't know about the UI method of installing/configuring.

teallarson (approved):

evantahler (approved):

evantahler:
> Looks good!
> 
> I wonder now that we have Config UI if we want to mention that and link to it somewhere in plugins docs? Maybe near the top? It's possible if someone lands on one of these plugin pages from a search that they wouldn't know about the UI method of installing/configuring.

I'll do that in a new PR - let's merge this one in!

PR #611, opened 2021-08-06
[Blog] - Node.js IPv4 + IPv6

evantahler:


pages/blog/node-js-and-ipv6.mdx
```diff
@@ -0,0 +1,108 @@
...
+});
+
+server.listen(port, hostname, () => {
+  console.log(`Server running at http://${hostname}:${port}/`);
+});
+```
+
+In this example, we are binding only to `127.0.0.1` which is the IPv4 version of what is called a `loopback` - this means that only the same computer can talk to itself. This is a very safe way to test and develop, and a very bad way to run a web server 🤣.
+
+Conversely, what if we wanted to allow traffic in from the widest variety of hosts? This is the default configuration of Grouparoo - we want the application to be as widely available as possible, and the infrastructure to be be in charge of restricting who the application can talk to as needed. In IPv4, that would mean choosing a host of `0.0.0.0` which would allow traffic from anywhere. What about IPv6? It turns out that the string `::` (yes, that’s to colons) means "all" in IPv6, and is actually shorthand for `0.0.0.0.0.0.0.0`.
```
andyjih:
```suggestion
Conversely, what if we wanted to allow traffic in from the widest variety of hosts? This is the default configuration of Grouparoo - we want the application to be as widely available as possible, and the infrastructure to be be in charge of restricting who the application can talk to as needed. In IPv4, that would mean choosing a host of `0.0.0.0` which would allow traffic from anywhere. What about IPv6? It turns out that the string `::` (yes, that’s two colons) means "all" in IPv6, and is actually shorthand for `0.0.0.0.0.0.0.0`.
```

pages/blog/node-js-and-ipv6.mdx
```diff
@@ -0,0 +1,108 @@
+---
+title: "Binding a Node.js Web app to IPv4 and IPv6 Sockets"
```
andyjih:
```suggestion
title: "Binding a Node.js web app to both IPv4 and IPv6 sockets"
```

pages/blog/node-js-and-ipv6.mdx
```diff
@@ -0,0 +1,108 @@
...
+date: "2021-08-05"
+author: "Evan Tahler"
+pullQuote: "How does Grouparoo use Next.js to load pages and components from plugins to modify our web user interface?"
```
andyjih: Is this pullQuote from another blog post?

andyjih (changes requested):
Looks good! Just some small stuff to tweak.

pages/blog/node-js-and-ipv6.mdx
evantahler: Whoops! Thanks!

```suggestion
pullQuote: "Grouparoo can speak IPv4 and IPv6 - Here's how we did it."
```

andyjih (approved):
Looks good! We'll keep it on hold for now.

PR #620, opened 2021-08-18
exclude trial_landing from sitemap

teallarson:
This addresses the recent Google Search Console error.


andyjih (approved):
Oops, thanks!

teallarson:
@andyjih all good!  We could look into another sitemap library that can look at metadata and just auto-exclude noindexed pages at some point... but this works for today.

PR #629, opened 2021-08-25
Calculated Property docs

teallarson:
Documentation for Calculated Property Plugin.  Located at `/docs/plugins/grouparoo-calculated-property` 

- [X] Generating an App
- [X] Generating a Source
- [X] Generating a Property
- [X] Examples of `customFunction`s in the config file


pages/docs/plugins/grouparoo-calculated-property.mdx
```diff
@@ -0,0 +1,24 @@
...
+pageNavSelector: "h2"
+---
+
+<PluginDocsLinks plugin="CalculatedProperty" />
+
+Grouparoo's Calculated Property plugin enables you to write a custom function to generate a new property based upon other existing Grouparoo Profile Properties.
+
+This guide will show you how to work with the Calculated Property plugin to [create a custom Calculated Property](#create-calculated-property-property)
+
+<PluginDocsInstallation plugin="CalculatedProperty" />
```
tealjulia, line 14: the casing gets weird in the code generated here because the slug uses hyphens, but the install template uses snake case... it bothers me, but I'm wondering if it's ok to others?

components/docs/plugins/createApp.tsx
```diff
@@ -9,9 +9,9 @@ import OptionsList from "./optionsList";
 const PluginDocsCreateApp = ({ plugin }: { plugin: string }) => {
   if (!plugin) return null;
 
-  const cmdName = plugin.toLowerCase();
-  const appId = `my_${cmdName}_app`;
   const pluginData = PluginData.find(({ name }) => name === plugin);
+  const cmdName = pluginData.slug;
+  const appId = `my_${cmdName.replace("-", "_")}_app`;
```
teallarson, line 14: This is our first plugin with a multi-word name!  To fit the casing conventions of our plugins/docs consistent, I added this.  Slugs and commands will use hyphens, object ids will remain snake case.

PR #633, opened 2021-08-30
add new plugins to Manifest and `/integrations/*`

teallarson:
- [X] Mailjet
- [X] Mixpanel
- [X] Calculated Properties (manifest only, not on integrations list)

Did not add individual integration pages for Mailjet or Mixpanel for this PR.


data/plugins/calculated-property.ts
```diff
@@ -0,0 +1,16 @@
...
+  priority: 0,
+  slug: "calculated-property",
+  logo: "calculated-property.png",
+  primaryType: "source",
+  otherTypes: [],
+  category: "transform tool",
+  badge: "",
+  showLink: false,
+  packageName: "@grouparoo/calculated-property",
+};
```
tealjulia, line 16: This file generates the manifest entry for calculated properties... it is not imported into the `/integrations` page at all but is still needed for the manifest.

data/plugins/index.ts
```diff
@@ -95,7 +97,9 @@ export const PluginData: Plugin[] = [
   Intercom,
   Iterable,
   Mailchimp,
+  Mailjet,
   Marketo,
+  Mixpanel,
```
tealjulia, line 102: Adds Mailjet and Mixpanel to the `/integrations` page

bleonard:
Is it true that we should wait to merge this until the 0.6 when the plugins are released? I think that's right because UI config reads the manifest.

teallarson:
> Is it true that we should wait to merge this until the 0.6 when the plugins are released? I think that's right because UI config reads the manifest.

@bleonard  Sounds right.  I just figured out why it was being erased from the manifest every time I pushed.  I'm thinking this commit should have it good to go!

andyjih (changes requested):
Looks good! The only note is that the calculated property image looks a bit skewed. 

Screen Shot 2021-08-31 at 9 49 27 AM

teallarson:
> Looks good! The only note is that the calculated property image looks a bit skewed.
> 
> Screen Shot 2021-08-31 at 9 49 27 AM

Thanks!  I think we actually don't want it to display calculated properties at all on this page.  I did make it square to display properly in UI config, but I also removed it from the `/integrations` page.

andyjih (approved):

PR #634, opened 2021-08-30
Added kc to employees list

krishnaglick:
Fighting with the image


components/home/whoWeAre.tsx
```diff
@@ -1,4 +1,4 @@
-import { Row, Col } from "react-bootstrap";
+import { Col, Row } from "react-bootstrap";
```
krishnaglick, line 1: Prettier

components/home/whoWeAre.tsx
```diff
@@ -196,6 +197,32 @@ export default function WhoWeAre() {
...
+                    target="_blank"
+                    rel="noreferrer nofollow"
+                  >
+                    @iHeartMedia
+                  </a>
+                </span>
+              }
+              github="https://github.com/krishnaglick"
+              linkedIn="https://twitter.com/msdatadetective"
+              twitter=""
```
krishnaglick: Should I make a twitter for myself as a grouparoo employee?

evantahler: If you don't have a Twitter account, no need to make a new one.  In that case, the component should be updated to handle a "" or null twitter account to hide the icon

components/home/whoWeAre.tsx
```diff
@@ -196,6 +197,32 @@ export default function WhoWeAre() {
...
+                    href="https://www.iheart.com"
+                    target="_blank"
+                    rel="noreferrer nofollow"
+                  >
+                    @iHeartMedia
+                  </a>
+                </span>
+              }
+              github="https://github.com/krishnaglick"
+              linkedIn="https://twitter.com/msdatadetective"
```
evantahler: I don't think that's you...

evantahler (changes requested):

components/home/whoWeAre.tsx
krishnaglick: Fixed!

PR #669, opened 2021-09-03
pass campaign as a prop to subscribe component

teallarson:
Subscribe component now accepts `campaign` as a prop, allowing us to customize further later on if desired.

### DIFF
--- a/components/subscribe.tsx
+++ b/components/subscribe.tsx
@@ -5,7 +5,7 @@ import { Row, Col, Form, Button } from "react-bootstrap";
 import { ErrorHandler } from "../utils/errorHandler";
 import { isBrowser } from "../utils/isBrowser";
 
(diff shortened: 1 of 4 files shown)


components/subscribe.tsx
```diff
@@ -5,7 +5,7 @@ import { Row, Col, Form, Button } from "react-bootstrap";
 import { ErrorHandler } from "../utils/errorHandler";
 import { isBrowser } from "../utils/isBrowser";
 
-export default function Subscribe() {
+export default function Subscribe(props) {
```
evantahler: This works, but I the props can be defined like this to ensure that the campaign name is always required when the component is used: 

```suggestion
export default function Subscribe({campaign}:{campaign:string}) {
```

PR #676, opened 2021-09-10
[whats-new] automatic profile deletion

pedroslopez:
We now support automatically clearing out profiles!

@evantahler should we start using the new "Records" language? Not sure when we want to start doing this.


evantahler (approved):

evantahler:
Short and sweet!

PR #685, opened 2021-09-30
add models to config ui docs

teallarson:
## Change description
Screen Shot 2021-10-05 at 12 28 47 PM

> Adds a Model page to UI Config docs as well as adding Models to the Core Product Topics.
> The Config UI doc for Models is where the setup step from Core will link to.

> Main changes found at: /docs/config/models

> NOTE: The "Core Concepts" description I left referring to the existing Profiles entry, but in the docs page, I've referred to Records instead.  Not sure which to go with as we have a separate story on doing a full docs update forthcoming.

## Related issues

Does this PR fix a Github Issue?
No

### Development

- [X] Application changes have been tested appropriately

### Impact

- [X] Code follows company security practices and guidelines
- [X] Security impact of change has been considered
- [X] Performance impact of change has been considered
- [X] Possible migration needs considered (model migrations, config migrations, etc.)

> n/a

### Code review

- [X] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [X] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)


data/docs-nav.ts
```diff
@@ -48,6 +48,7 @@ const DocsNav: NavItem[] = [
     path: "/docs/config",
     children: [
       { title: "Apps", path: "/docs/config/apps" },
+      { title: "Models", path: "/docs/config/models" },
```
evantahler, line 51: I think models should come before apps maybe?

pages/docs/config/models.mdx
```diff
@@ -0,0 +1,45 @@
...
+title: Models
+date: "2021-06-22"
+pullQuote: "Learn how to configure apps using Grouparoo's UI Config."
+---
+
+Models are how you define the shape you want your records to have in Grouparoo. Once you've created a model, you can add sources and properties to that model to generate different record types. For example, you may want one model that defines leads, another for customers, and another for site visitors. [See this doc](/docs/getting-started/product-concepts) to learn more about Grouparoo's core concepts.
+
+With [UI Config](/docs/config) running and registered, follow the steps below to configure your model(s).
+
+### Add a Model
```
evantahler: The other pages have a "Step 1"

pages/docs/config/models.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+<ImageInBrowserFrame
+  centered
+  alt="UI Config: Add Model"
+  src="/images/docs/guides/config/model--01--add.png"
+  width={820}
+  height={402}
+/>
+
+You will see that Grouparoo automatically generates a Profile model for you. You can add a new model by clicking "Add Model".
```
evantahler: https://www.pivotaltracker.com/story/show/179745072 means we will not create a model automatically any more - might as well leave this out

pages/docs/config/models.mdx
```diff
@@ -0,0 +1,45 @@
...
+Next you can give your model a name. For now, that's all you need to do!
+
+<ImageInBrowserFrame
+  centered
+  alt="UI Config: Add Model"
+  src="/images/docs/guides/config/model--03-name.png"
+  width={820}
+  height={399}
+/>
+
```
evantahler, line 42: Perhaps:

> In the future, Grouparoo will be adding more types of models to support additional types of Properties and Destinations.

evantahler (approved):
Approved, pending some minor content changes.

I agree that the later stories are the ones in which we update "Core Concepts" and the reset of the site/docs.

data/docs-nav.ts
teallarson: Should it?  I was trying to think in terms of hierarchy -- an App can have more than one Model, but a Model can only be associated with one App, right?  I was also thinking in terms of the order they'll appear in the UI navbar.  Happy to make the switch, I'd just like to understand more!

pages/docs/config/models.mdx
teallarson: It's so awkward without a "Step 2"!  I'll add it for consistency.

data/docs-nav.ts
evantahler: This is a good taxonomy question!  Both being "first" is possible because Apps and and Models don't meet up logically until Sources.

```
App    ↘
         Source -> Property -> Group... 
Model  ↗ 
```

* You could have one app (datawarehouse) that has many models (customer, lead, events).  App -> Model
* You could have many apps that each contribute parts to many models (snowflake has both the "users" table for customers and leads; redshift has events that are used for both customers and leads to calculate LTV).   Model -> App

After writing this out, I think you have it right initially - let's continue to talk about Apps first because that's probably how data-engineers think - I've got a datawehouse full of tables, not a collection of models that happen to be in a datawarehouse.

PR #722, opened 2021-10-15
NextJS Blog Post

krishnaglick:


data/authors.ts
```diff
@@ -3,7 +3,7 @@ export interface Author {
   slug: string;
   image: string;
   url: string;
-  twitter: string;
+  twitter?: string;
```
evantahler, line 6: 👍

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+With the preamble out of the way, here's the meat of it. Since the application was rendered from a single point all application data was stored in a single, top level MobX store. This was dehydrated, shipped to the client, and then hydrated for application parity. The core concept is a single entrypoint always, and from there make decisions.
+
+That concept also influenced later work I did that was client-side only. Utilizing Create-React-App (CRA) I built a few contexts and shared them at top level effectively giving me multiple stores to pull and interact with. I quite liked this pattern as it let's you isolate your concerns without too much of a headache and React hooks let you do things requiring data from more than one without too much trouble.
+
+```javascript
+import React, { useState } from "react";
+
+function Example() {
+  // Declare a new state variable, which we'll call "count"  const [count, setCount] = useState(0);
```
evantahler: Why is this commented out, and not doing the `useState()`?

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,99 @@
...
+```
+
+```tsx
+export const getServerSideProps: GetServerSideProps<MyProps> = async (
+  context
+) => {
+  const data = await api.getMyData(context.req);
+  return { props: { data } };
+};
+```
```
evantahler: So is this part of the example of what I should do or what I shouldn't do?

evantahler (changes requested):
I like the intro giving some personal and historic context!  I think you may have buried the lede a bit though.  All the way at the bottom is what the blog post is actually talking about - how Grouparoo uses Next's page hydration, and how that separate server-side contexts don't match with client contexts.  I'd suggest starting with an intro about the problem, and then getting into the history.

Your examples are good, but they are lacking context:
* Are these examples the problem code or the solution code?
* Where are these files?  Are they next,js pages? 

I think I'd want the blog post to end with an example that does work and shows how I could hydrate a context from data given to the page in `getInitialProps()`

Also, you may wan to link out to the next.js docs so people can learn more about why Next is different and why using getInitialProps is helpful

pages/blog/2021-10-14-nextjs-and-data.mdx
krishnaglick: I was throwing together some React docs stuff, plus modifying some code we have. I can improve it for sure.

pages/blog/2021-10-14-nextjs-and-data.mdx
krishnaglick: Copied react docs examples.

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
+---
+title: "NextJS and Managing Your Data"
+date: "2021-10-15"
+tags: ["engineering"]
+description: "Lessons learned trying to apply single-store concepts to a NextJS app"
+author: "Krishna Glick"
+pullQuote: "NextJS is not a single point of entry application."
+---
+
+Managing data in a front-end framework is an infinitely solved problem. Every framework has its own flavor, architecture, and opinions. Notably, NextJS makes does not. They provide a variety of methods for [data fetching](https://nextjs.org/docs/basic-features/data-fetching) but offers no in-built patterns for data management.
```
evantahler: `Notably, NextJS makes does not.` - not what?  Maybe `Notably, NextJS does not have it's own opinions and leaves that up to you and React.`

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+On one hand, it doesn't seem unreasonable; NextJS simply expands on React. On the other, that leaves it up to the individual or team to make those choices. React Contexts are fairly powerful, though limited during server-side rendering. Global stores like React Tooltik, Recoil, or Mobx are all more complete solutions with their own pros and cons.
+
+I ultimately went with React Contexts given the use-case of the application I was building here at Grouparoo.
```
evantahler: ... and they work on both the server and browser in some magic way?  They don't conflict?  Contexts know how to get out of the way on the server and be present for the client?

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+For some background, I was a Lead Engineer at iHeartMedia on a very high-traffic application. If you've ever visited a website for a radio station under the iHeart brand you've seen the result of my teams engineering efforts. That application started development in late 2018 with one goal at the forefront: let React handle complex, branching async rendering, and async data loading.
```
evantahler: of my team's (apostrophe) engineering efforts.

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+<Image
+  alt="iHeart wmxw-fm"
+  src="nextjs-and-data/iheart-react.png"
+  width="700"
+  height="333"
+/>
+
+As the application rendered each portion of the tree it would need to fetch additional data to know what portions to render next. NextJS only offered async rendering at the top of the tree, and at the time the React team claimed it would have functional server-side rendering out soon. As of this writing it's still [not available](https://github.com/reactwg/react-18/discussions/37_) in the form that was promised.
```
evantahler: https://github.com/reactwg/react-18/discussions/37_ - extra `_`

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+Since the application was rendered from a single point all application data was stored in a top level MobX store. This was dehydrated, shipped to the client, and then hydrated for application parity. The core concept is a single entrypoint always, and from there make decisions.
```
evantahler: Maybe a link to https://en.wikipedia.org/wiki/Hydration_(web_development)

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
...
+export const Some = () => {
+  const { someValue } = useContextAContext();
+
+  return <div>{someValue}</div>;
+};
+```
+
+At Grouparoo we needed to build a UI for one of our internal applications, and thanks to [ActionHero's NextJS Plugin](https://github.com/actionhero/ah-next-plugin) we slid a new UI into place quite easily. I approached NextJS like I would a greenfield CRA - creating some top level contexts and pushing my data into them. This ran into problems quickly.
+
+As seen above, I like to make use of React hooks extensively. Since hooks don't run on the server any logic I built to load and store data had to _also_ happen in a `getServerSideProps`, or other NextJS hook.
```
evantahler: > Since hooks don't run on the server, any logic I built (comma)

pages/blog/2021-10-14-nextjs-and-data.mdx
```diff
@@ -0,0 +1,109 @@
...
+export default App;
+
+//contextA.ts
+export const useContextAInitialState = ({ data }: { data: string }) => {
+  const [someValue, updateSomeValue] = useState(data || "someValue");
+  return { someValue, updateSomeValue };
+};
+```
+
+The other concept I utilized was putting _all_ data into [contexts](https://reactjs.org/docs/context.html). With a client-only application it made sense; going to a new "page" required limited to no additional data. In NextJS land this didn't work so well. Since each page has its own data-loading and context data was set at top level it required some gymnastics to get the page-loaded data into the React contexts.
```
evantahler: had utilized in the past?

evantahler (approved):
Much clearer narrative!

This is ready-to-go as-is, but I've added some gramatical notes below

pages/blog/2021-10-14-nextjs-and-data.mdx
krishnaglick: Since I go into more detail later, I think nuking this one line is good.

pages/blog/2021-10-14-nextjs-and-data.mdx
krishnaglick: I've rephrased, lemme know what you think!

evantahler (approved):
Looks great! Ship it!

PR #747, opened 2021-10-28
Change default test time from 5 seconds to 10

bleonard:
CI was timing out frequently.


evantahler (approved):
that's a shame, but sure!

I think that CircleCI is having a bad/slow week too

PR #818, opened 2021-12-07
Enable Eslint

krishnaglick:
## Change description

Eslint is enabled, and things are fixed.

### Development

- [ ] Application changes have been tested appropriately

### Impact

- [ ] Code follows company security practices and guidelines
- [ ] Security impact of change has been considered
- [ ] Performance impact of change has been considered
- [ ] Possible migration needs considered (model migrations, config migrations, etc.)

### Code review

- [ ] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [ ] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)


components/DebugImage.tsx
```diff
@@ -23,25 +23,28 @@ export default function DebugImage(props) {
...
+      setData({ error, message });
+    }
+  }, [props.height, props.src, props.width]);
+
+  useEffect(() => {
+    if (!isDebugging(props)) {
+      fetchData();
+    }
+  }, [fetchData, props]);
+
```
krishnaglick, line 44: It looked like we didn't want to fetch data if we were debugging, so I've moved that logic into the hook. React always wants hooks to run if they're declared.

components/blog/Image.tsx
```diff
@@ -35,7 +35,7 @@ function BlogImage(props: BlogImageProps) {
 
   return (
     <div style={containerStyle}>
-      <Image {...imageProps} />
+      <Image alt={imageProps.alt} {...imageProps} />
```
krishnaglick, line 38: The linting was getting caught up on this, since it wasn't explicitly set.

components/docs/Image.tsx
```diff
@@ -2,19 +2,18 @@ import Image, { MyImageProps } from "../Image";
...
 
   if (
-    src.includes("data:image") ||
-    src.startsWith("http:") ||
-    src.startsWith("https:")
+    !(
+      src.includes("data:image") ||
+      src.startsWith("http:") ||
+      src.startsWith("https:")
+    )
```
krishnaglick, line 13: Removing some redundancy.

components/docs/Image.tsx
```diff
@@ -2,19 +2,18 @@ import Image, { MyImageProps } from "../Image";
...
   ) {
-    imageProps.src = src;
-  } else {
     imageProps.src = `/images/${src}`;
   }
 
-  return <Image className="mb-4" {...imageProps} />;
+  return <Image className="mb-4" {...imageProps} alt={imageProps.alt} />;
```
krishnaglick: Explicit alt tag :|

components/layouts/Floating.tsx
```diff
@@ -103,7 +103,6 @@ function FloatingLayout({ children }) {
           content="/favicon/mstile-310x310.png"
         />
         <meta name="viewport" content="width=device-width" />
-        <link rel="stylesheet" href="/css/prism.css"></link>
```
krishnaglick, line 106: Moved this into `grouparoo.scss`

pages/chat.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @next/next/no-sync-scripts */
+/* eslint-disable react/no-unescaped-entities */
```
krishnaglick, line 2: Funky page, better to just disable

pages/docs/config/code-config.mdx
```diff
@@ -78,7 +78,7 @@ If you want to set a variable called `production_hubspot_api_key` for use only i
 
 <Alert variant="warning">
   <p className="mb-0">
-    ⚠️ <code>APP</code> (the object type) is surrounded by double underscores.
```
krishnaglick: This was a weird unicode character lying around.

components/banners/WhatsNewBanner.tsx
```diff
@@ -13,7 +13,7 @@ export function WhatsNewBanner({
   return (
     <Alert variant="primary" style={{ textAlign: "center", margin: 0 }}>
       <small>
-        <strong>What's New</strong> - {releaseNote.description}{" "}
+        <strong>What&apos;s New</strong> - {releaseNote.description}{" "}
```
evantahler: I'd probably go with quoted strings here `{"what's new"}` vs html codes

evantahler: Personally, I find this kind of annoying and would rather turn this rule off - TS compiling helps know if our strings are terminated or not

evantahler (changes requested):
I'm ok with most of these changes, except the rule that messes with our quoting and requires HTML codes.  This is a documentation site, and if we can't easily write words without quoting them in special ways, I'm guessing that it would be annoying.

Do you need help fixing the test?

components/banners/WhatsNewBanner.tsx
krishnaglick: Happy to turn it off!

pages/docs/config/code-config.mdx
evantahler: I think we really do want to show this emoji:

Screen Shot 2021-12-09 at 4 28 48 PM

pages/index.tsx
```diff
@@ -25,7 +25,7 @@ export default function IndexPage({ pageProps, setReleaseNote }) {
     pluginNames[Math.floor(Math.random() * pluginNames.length)];
   const tagline = `Stop writing code to sync data to ${pluginName}*`;
 
-  useEffect(() => setReleaseNote(releaseNote), []);
+  useEffect(() => setReleaseNote(releaseNote), [releaseNote, setReleaseNote]);
```
evantahler, line 28: I don't think I understand re-triggering the useEffect hook when the setter (`setReleaseNote`) changes - what would cause that?

components/solutions/ComparisonPage.tsx
```diff
@@ -88,5 +88,4 @@ export default function ComparisonPage(comp: ComparisonInfo) {
       <ComparisonTable comparisonChartData={comp.comparisonChartData} />
     </>
   );
-  return <>"abc"</>;
```
evantahler, line 91: whoops!

components/Image.tsx
```diff
@@ -101,4 +102,6 @@ const MyImage = React.forwardRef<HTMLImageElement, MyImageProps>(
   }
 );
 
+MyImage.displayName = "MyImage";
+
```
evantahler, line 106: What's this do?

components/docs/plugins/CodeBlock.tsx
```diff
@@ -32,9 +32,9 @@ const PluginDocsCodeBlock = ({ cli, code, language }: CodeBlockProps) => {
       {({ tokens, getLineProps, getTokenProps }) => (
         <pre className={`language-${language}`}>
           {tokens.map((line, i) => (
-            <div {...getLineProps({ line, key: i })}>
+            <div {...getLineProps({ line, key: i })} key={i}>
               {line.map((token, key) => (
-                <span {...getTokenProps({ token, key })} />
+                <span {...getTokenProps({ token, key })} key={key} />
```
evantahler, line 37: These don't look right any more
Screen Shot 2021-12-09 at 4 34 59 PM

evantahler (changes requested):

pages/index.tsx
krishnaglick: Because the function comes from props, the `react-hooks/exhaustive-deps` says it should be in there.

components/Image.tsx
krishnaglick: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md

krishnaglick: I'd be willing to disable this one, but it seems like it couldn't hurt.

components/docs/plugins/CodeBlock.tsx
krishnaglick: Sass importing css can get funky, but should be fixed now.

evantahler (approved):
🥳

PR #913, opened 2022-02-01
Replace `grouparoo generate` with `grouparoo config` in CLI docs

edmundito:
## Change description

This change removes the references to `grouparoo generate` from CLI docs and replaces them with `grouparoo config` instead.

### Development

- [x] Application changes have been tested appropriately

### Impact

- [x] Code follows company security practices and guidelines
- [x] Security impact of change has been considered
- [x] Performance impact of change has been considered
- [x] Possible migration needs considered (model migrations, config migrations, etc.)

### Code review

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [x] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)

### DIFF
--- a/components/docs/CLICommands.tsx
+++ b/components/docs/CLICommands.tsx
@@ -1,7 +1,11 @@
 import { Fragment } from "react";
 import CLICommandData from "../../data/cli-commands.json";
 
-export default function CLICommand({ name }) {
+interface Props {
+  name: string;
+}
+
+export default function CLICommand({ name }: Props) {
   const data = CLICommandData.find((c) => {
     const chunk = c.name.split(" ")[0];
     return chunk === name;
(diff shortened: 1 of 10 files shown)


teallarson (approved):
Looks good!  

Likely a different PR: Do you think it's confusing that we use "code config" and "config files" across the docs somewhat interchangeably?  I don't know that it's spelled out that they're referring to the same thing.

edmundito:
@tealjulia Per a conversation in Slack, I pushed another update to hide `code-config` and replaced references in the docs (excluding blog, plugins, and integrations) to mention config UI instead. The reason it's just hidden is that a lot more work is needed in how these docs should be written so this at least removes them from the common navigation.

> Do you think it's confusing that we use "code config" and "config files" across the docs somewhat interchangeably?

That's a larger question on whether we should talk about the config files. But definitely editing the config files and code config are separate things. I decided to only focus on replacing code config references with running the UI and resolving this separately.

PR #918, opened 2022-02-08
Add Configuration Files support docs

edmundito:
## Change description

This adds a Configuration Files doc to the support section describing the details of each configuration file. The content is based on the code-config docs with additional information and without references to the generators.

### Development

- [x] Application changes have been tested appropriately

### Impact

- [ ] Code follows company security practices and guidelines
- [ ] Security impact of change has been considered
- [ ] Performance impact of change has been considered
- [ ] Possible migration needs considered (model migrations, config migrations, etc.)

### Code review

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [x] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)


pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
+---
+title: "Configuration Files"
+date: "2022-02-09"
+pullQuote: "Definitions and examples of Grouparoo configuration files."
+---
+
+Grouparoo projects contain configuration files for Apps, Models, Sources, Properties, Groups, Destinations, and any other objects is exported into config files.
+These config files can be created and updated with [UI config](/docs/config).
+
+Config files are either JSON or JavaScript which allow you the flexibility to manipulate data or load configuration from external sources.
```
edmundito: I left references to the `JavaScript` files from code config. But should this be removed? In a world without generators should we only encourage JSON files?

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
+---
+title: "Configuration Files"
+date: "2022-02-09"
+pullQuote: "Definitions and examples of Grouparoo configuration files."
+---
+
+Grouparoo projects contain configuration files for Apps, Models, Sources, Properties, Groups, Destinations, and any other objects is exported into config files.
```
Adorism:
```suggestion
Grouparoo projects contain configuration files for Apps, Models, Sources, Properties, Groups, Destinations, and any other objects generated in the course of configuration. 
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+  - Must be unique among other objects of the same type
+- `class`: (**required**) The type of object that is being defined such as `App`, `Model`, `Source`, `Property`, `Group`, `Destination`
+- `name`: (**required**) The display name of this object
+
+## App Config
+
+The App configures the connections available to create Sources and Destinations.
+
+- `id`: (**required**) Unique identifier for this app config.
+- `name`: (**required**) A display name. It matches the `id` by default.
```
Adorism: Am I right to think that if the name is assigned in config UI and does not meet the requirements for ID, the name will be altered to assign a valid id? If so, do you think we should we add something like this so users aren't confused about rules for assigning a name in the UI if they read this?

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+- `class`: (**required**) This must be set to `"App"`.
+- `type`: (**required**) The type of the app define by a plugin. While the type can match the name of the plugin, plugins may offer multiple apps and the names may not match.
```
Adorism:
```suggestion
- `type`: (**required**) The type of the app defined by a plugin. While the type can match the name of the plugin, plugins may offer multiple apps and the names may not match.
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+
+- BigQuery
+- ClickHouse
+- MySQL
+- Postgres
+- Redshift
+- Snowflake
+- SQLite
+
+For the best results, configure your query up to return a single row of data. For example, you might set it up as:
```
Adorism:
```suggestion
For the best results, configure your query to return a single row of data. For example, you might set it up as:
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+  "class": "Model",
+  "id": "users",
+  "name": "Users",
+  "type": "profile"
+}
+```
+
+## Source Config
+
+Each Source has a common set of options, while other options are specific to the _type_ of Source. Here are the common options:
```
Adorism:
```suggestion
Each Source has a common set of options. Certain _types_ of Sources may have additional options, but all Sources will have the following common options:
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+    }
+  ]
+}
+```
+
+If you've configured an App Refresh Query when you configured your App, you can also opt to leave a particular Schedule out of its updates by adding `"refreshEnabled": false` to the Schedule's config.
+
+## Property Config
+
+Each Property has a common set of options, while other options are specific to the _type_ of Property. Here are the common attributes:
```
Adorism:
```suggestion
Each Property has a common set of options. Certain _types_ of Properties may have additional options, but all Properties will have the following common options:
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+
+Available operations depends on both the _type_ of Property and the underlying Grouparoo database (SQLite vs. Postgres). See below for a list of rules. (If in doubt, choose Postgres.)
+
+Note you can use SQL wildcard characters (%, \_, [ ], ^, -) in your Group rule match values as appropriate and necessary.
+
+<RuleOpsTable />
+
+## Destination Config
+
+Each Destination has a common set of options, while other options are specific to the _type_ of Destination. Here are the commons set of options that may look similar to other config object shapes in Grouparoo:
```
Adorism:
```suggestion
Each Destination has a common set of options. Certain _types_ of Destinations may have additional options, but all Destinations will have the common options. These common options, which may look very similar to other config object shapes in Grouparoo, are listed here: 
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+    "LNAME": "lastName"
+  }
+}
+```
+
+Notice the odd structure of the keys. That's specific to Mailchimp.
+
+**Destination Group Membership**
+
+Grouparoo also provides the ability to attach Group membership to the exported data. How this looks within the Destination is specific to the Destination. But the shape of the config file largely remains the same. In most Destinations, Destinations Group Memberships become either tags (Mailchimp), lists (Hubspot), or appended as additional properties to the Record (Customer.io).
```
Adorism:
```suggestion
Grouparoo also provides the ability to attach Group membership to the exported data. How this looks within the Destination is specific to the Destination. But the shape of the config file largely remains the same. In most Destinations, Destination Group Memberships either become tags (Mailchimp) or lists (Hubspot), or are appended as additional properties to the Record (Customer.io).
```

pages/docs/support/config-files.mdx
```diff
@@ -0,0 +1,672 @@
...
+
+```bash
+export GROUPAROO_OPTION__{topic}__{name}={value}
+```
+
+See [the Secrets page](/docs/deployment/secrets) for more information about using environment variables for config.
+
+**JavaScript Functions**
+
+Code config files can be use JavaScript functions that can export an `async` method, which can `await` while loading data from an database or API. (See the _Examples_ section below.)
```
Adorism:
```suggestion
Code config files can use JavaScript functions that can export an `async` method, which can `await` while loading data from a database or API. (See the _Examples_ section below.)
```

pages/docs/support/config-files.mdx
edmundito: I'll remove this because it's related to the generator. In config UI, the name is driven by what the user types and this is more related to how the generator worked

edmundito:
@Adorism I moved the advanced usage cases to its own file instead and reworded things a bit.

Adorism (approved):

